PR #1, opened 2021-05-18
Filename issues

SepiDre:
I got an huge issue with not supported chars in Windows, so I removed them.


marzzzello:
Yes, that's Windows :D
If you add `if os.name == "nt":`, I'll merge it
Thanks!

PR #22, opened 2022-02-09
added feature for auto calculating days since last download

kjabl:
Using pytr I recognized it's always downloading all files regardless wether there are already files present in the target directory. So you get a lot of duplicate files if you don't set a proper --last_days parameter. One way is to calculate this day by hand and enter it or use a shell script to call pytr with a proper parameter.
For myself I thought it would be nice feature if pytr could do this automatically. So I decided to add this feature and thought this could be interesting for others, too.

What does my modification do?
I added a new parameter. Calling pytr with --auto_days specified leads to --last_days being ignored but checking all .pdf files in the target directory for the most recent date. I used st_mtime which is the last modication time. I didn't use st_ctime because on unix systems this can be updated through a metadata change.
After finding the most recent date it calculates the number of days between then and now and sets this as a value for --last_days respective since_timestamp
The calculation is done in utils.py.
If the difference is 0 it rejects execution as 0 would mean no limit. So this can only be used if you didn't download anything the same day.

I know it would be better to check the files for duplicates but this would require downloading them first. Another reason I didn't go for that approach is that I'm not sure about how to check to files for equality as I don't know what happens to document titles and so on if you purchase the same stock twice a day for example. If there is any information about this I would be happy to start another more precise approach.

In the end I would be happy if you consider this a useful addition and accept the pull request :)

### DIFF
--- a/pytr/main.py
+++ b/pytr/main.py
@@ -9,7 +9,7 @@ import shtab
 
 from importlib.metadata import version
 
-from pytr.utils import get_logger, check_version
+from pytr.utils import get_logger, check_version, auto_calculate_since_timestamp
 from pytr.dl import DL
 from pytr.account import login
 from pytr.portfolio import Portfolio
@@ -62,6 +62,9 @@ def get_main_parser():
     )
     parser_dl_docs.add_argument(
         '--last_days', help='Number of last days to include (use 0 get all days)', metavar='DAYS', default=0, type=int
+    )
+    parser_dl_docs.add_argument(
+        '--auto_days', help='Auto calculate days since previous download from existing files (ignores --last_days)', action='store_true'
     )
 
     parser_get_price_alarms = subparsers.add_parser(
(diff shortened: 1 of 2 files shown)


networxnet:
I got no duplicates 🤔

kjabl:
Oh thats interesting. I checked again and it seems I was kinda wrong. Not every file is duplicated but some are. 
If I run
`pytr dl_docs .` twice some files will be duplicates. After first run I had 113 files in "Sparplan/Abrechnung Ausführung" and in total 398 files. After second run I had 173 in "Sparplan/Abrechnung Ausführung" and 483 in total. So most of the duplicates seems to be in that folder.
The duplicate even gets a slightly different name but the content is exactly the same.
grafik

networxnet:
So.it ist a Problem in TR side.

Correct me but I think it is difficult to solve this as this are two different files ...

kjabl:
Well the content is exactly the same but my approach isn't the right one I guess.
The only difference I see is the filename, which is build using the format string (in my case the default)
But I didn't dive into it deep enough to judge where the different names come from. So I think you should deny this pull request.
I'll observe this behavior in the future. If I find something that could be fixed I'll let you know.

marzzzello:
If the file is already present, it won't get re-downloaded:

https://github.com/marzzzello/pytr/blob/ed94ff99189301a8bdd5033f59638977e306336b/pytr/dl.py#L100-L112

marzzzello:
I found out what the problem is:
On the first run, the downloader checks if the filename already exists and also checks if the URL is in the downloader queue.
So if the filename is different but the URL is the same, the file will get skipped.
https://github.com/marzzzello/pytr/blob/ed94ff99189301a8bdd5033f59638977e306336b/pytr/dl.py#L103

On the second run the URL is not in the queue because is got skipped as the filename already exists. But the other document with the different title does not exist as the filename is different. And the URL is not in the queue so it gets downloaded

kjabl:
If I understand that correctly we have one single URL which can serve documents with the same content but different metadata (title)?
That sounds realy difficult/impossible to solve. Perhaps my approach is not that wrong in this case? It's got negative aspects though. If you download run in the morning and some days later run again, documents published in the evening after the first wont bei downloaded for example.

The only other way I see at the moment would be to check the content for equality. But that would mean a deeper analysis and accordingly slower performance :/

kjabl:
I just added a different PR https://github.com/marzzzello/pytr/pull/23 . Checking MD5 hash of PDF files to avoid these duplicate. In my opinion that would be the better solution. And it is definitely working for me!

PR #23, opened 2022-02-11
added ability to check hashes of downloaded PDF-files to avoid duplic…

kjabl:
As discussed in https://github.com/marzzzello/pytr/pull/22 there can be duplicates download files if you run more then once.

To avoid this I added an ability to check md5 hashes. The downside is you need to download the file first, check and delete. So I didn't add it as default. You need to enable it by using --check_hashes when callin pytr.

For me (400 documents) I didn't recognize a slower performance while saving to SSD drive. But I could be different with more documents I guess.

### DIFF
--- a/pytr/dl.py
+++ b/pytr/dl.py
@@ -1,3 +1,4 @@
+import os
 import re
 
 from concurrent.futures import as_completed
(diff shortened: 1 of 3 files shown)


networxnet:
You put this into 0.19 soon @marzzzello ?

marzzzello:
Closed in favor of https://github.com/marzzzello/pytr/commit/7bd44ca555b91592f90a0eb29945f9ca7d24fa00

PR #29, opened 2022-04-22
If multiple documents would have the same file path when downloading,…

britiscurious:
… try to append the document id to make the file path unique.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ def readme():
 
 setup(
     name='pytr',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -10,7 +10,7 @@ def readme():
 
 setup(
     name='pytr',
-    version='0.1.5',
+    version='0.1.7',
```
miry, line 13: I think it should be `0.1.6`

PR #50, opened 2024-02-02
feat: add transaction export to YNAB import format

morremeyer:
This adds a `transactions` command that is similar to the `export_transactions` command.

However, `transactions` uses the `timelineTransactions`, which contain all transactions and exports the JSON as seen on the websocket to a target file.


pytr/transactions.py
```diff
@@ -0,0 +1,48 @@
...
+                t = self.transactions[-1]
+                if datetime.fromisoformat(t['timestamp']) < datetime.now().astimezone() - timedelta(days=self.last_days):
+                    return
+
+                await self.tr.timeline_transactions(response["cursors"]["after"])
+
+            else:
+                print(f"unmatched subscription of type '{subscription['type']}':\n{preview(response)}")
+
+            if recv == 1:
```
Katzmann1983, line 33: The recv value is not updated in the while loop, or is it?

pytr/main.py
```diff
@@ -155,6 +156,20 @@ def formatter(prog):
...
+    info = 'Export all transactions to csv'
+    transactions = parser_cmd.add_parser(
+        'transactions',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+        parents=[parser_login_args],
+        help=info,
+        description=info,
+    )
+    transactions.add_argument(
+        'output', help='Output path of CSV file', metavar='OUTPUT', type=Path)
```
Katzmann1983, line 169: For some commands output is a file, sometimes a folder. That may be confusing, no?

pytr/transactions.py
```diff
@@ -0,0 +1,48 @@
...
+                return
+
+    def output(self):
+        transactions = [
+            t
+            for t in self.transactions
+            if datetime.fromisoformat(t['timestamp']) > datetime.now().astimezone() - timedelta(days=self.last_days)
```
Katzmann1983, line 40: just calc once and set as internal variable of the transaction:
self.since_timestamp

pytr/api.py
```diff
@@ -396,6 +396,13 @@ async def portfolio(self):
     async def portfolio_status(self):
         return await self.subscribe({'type': 'portfolioStatus'})
 
+    async def timeline_transactions(self, after=None):
```
Katzmann1983, line 399: This function already exists in the api.py, no?

morremeyer:
Thanks for your feedback, I'll address it in the upcoming days!

pytr/api.py
morremeyer: Yes, it was added by e4d1fc7ed22770b8aa9dbffaee3417de75aedc2b, which was merged after I initially opened the PR. 

I removed this implementation in favor of the already merged one.

morremeyer (commented):
@Katzmann1983 I added all your suggestions, thanks for the detailed review!

Sorry for the force push here, I only realized once I had already pushed. 

My only question is in regards to the output parameter, I don't think I quite got your comment there.

morremeyer:
Force pushed to include all the linting with black, and other changes that happened since the last review.

pytr/transactions.py
```diff
@@ -57,3 +61,42 @@ def export_transactions(input_path, output_path, lang="auto", sort=False, date_i
         f.write(lines)
 
     log.info("Deposit creation finished!")
+
+
+class Transactions:
```
NiklasRosenstein, line 66:
```suggestion
class TransactionsJsonExporter:
```

NiklasRosenstein (changes requested):

pytr/transactions.py
```diff
@@ -57,3 +61,42 @@ def export_transactions(input_path, output_path, lang="auto", sort=False, date_i
...
+
+            if subscription["type"] == "timelineTransactions":
+                self.transactions.extend(response["items"])
+
+                # Transactions in the response are ordered from newest to oldest
+                # If the oldest (= last) transaction is older than what we want, exit the loop
+                t = self.transactions[-1]
+                if datetime.fromisoformat(t["timestamp"]) < self.not_before:
+                    return
+
```
NiklasRosenstein, line 86: I've tested this locally, and if you go far back enough to include the first transaction, it just keeps going over the timeline forever because there is no "after" after the chunk with the first transaction is returned.

```suggestion
                if not response["cursors"]["after"]:
                    break

```

pytr/main.py
```diff
@@ -206,6 +206,24 @@ def formatter(prog):
...
+        parents=[parser_login_args],
+        help=info,
+        description=info,
+    )
+    transactions.add_argument("output", help="Output path of JSON file", metavar="OUTPUT", type=Path)
+    transactions.add_argument(
+        "--last_days",
+        help="Number of last days to include (use 0 get all days, defaults to 14)",
+        metavar="DAYS",
+        default=14,
```
NiklasRosenstein, line 224: Seems like an unintuitive default to me, how about we make the default such that it returns all transactions?

pytr/main.py
```diff
@@ -206,6 +206,24 @@ def formatter(prog):
...
+    # transactions
+    info = "Export all transactions directly from the API to a file (JSON)"
+    transactions = parser_cmd.add_parser(
+        "transactions",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+        parents=[parser_login_args],
+        help=info,
+        description=info,
+    )
+    transactions.add_argument("output", help="Output path of JSON file", metavar="OUTPUT", type=Path)
```
NiklasRosenstein, line 219:
```suggestion
    transactions.add_argument(
        "output",
        help="Output path of JSON file [default: - (stdout)]",
        metavar="OUTPUT",
        type=argparse.FileType("w"),
        default="-",
        nargs="?",
    )
```

This would make it so if you give no arguments, you'd write to stdout. `TransactionsJsonExporter` would need to be updated to take a file-like object instead of a Path.

pytr/transactions.py
```diff
@@ -57,3 +61,42 @@ def export_transactions(input_path, output_path, lang="auto", sort=False, date_i
         f.write(lines)
 
     log.info("Deposit creation finished!")
+
+
+class TransactionsJsonExporter:
```
NiklasRosenstein, line 66: I think this class could serve better as a `TransactionsFetcher` or vice versa, it's only responsibility being what it currently does in `loop()`, and serve the fetched transactions back to the caller (e.g. via callbacks, (async) iterator, etc).

Dumping that data into the file as JSON is trivial and can easily be handled in `main.py`

pytr/main.py
NiklasRosenstein: Seems OK to me as well, the fact that it is supposed to be a file is in the help string.

NiklasRosenstein (changes requested):

pytr/dl.py
```diff
@@ -44,7 +44,7 @@ def __init__(
         self.filepaths = []
         self.doc_urls = []
         self.doc_urls_history = []
-        self.tl = Timeline(self.tr, self.since_timestamp)
+        self.tl = Timeline(self.tr, self.not_before)
```
NiklasRosenstein, line 47: Further testing on this branch reveals that this is causing an error because the argument is expected to be a `float` and not a `datetime` object.

```
  File "/home/niklas/git/pytr/pytr/timeline.py", line 42, in get_next_timeline_transactions
    or datetime.fromisoformat(event["timestamp"][:19]).timestamp() >= self.max_age_timestamp
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: '>=' not supported between instances of 'float' and 'datetime.datetime'
```

pytr/transactions.py
NiklasRosenstein: Committed this change to unblock request in #173

pytr/dl.py
morremeyer: Interesting, I verified that it was working before requesting your review again. 

Will check it out, thanks for the thorough review!

morremeyer: I finally got time to look into this again, and rebased the branch.

I fixed the issue in 717cee857af22b720f6ebabfcc64ad62cc6e1ab6.

pytr/main.py
```diff
@@ -320,6 +345,14 @@ def main():
     if args.verbosity.upper() == "DEBUG":
         log.debug("logging is set to debug")
 
+    # Compute the latest timestamp to get data for if last_days is specified
+    if hasattr(args, "last_days") and args.last_days > 0:
+        not_before = datetime.now().astimezone() - timedelta(days=args.last_days)
+    elif args.last_days == 0:
+        not_before = datetime.min.replace(tzinfo=timezone.utc) # Get everything
```
RealCLanger, line 352: Why this case and the else case? I guess we can chose either.

RealCLanger (commented):
After my latest changes, can you merge the PR with master? Then I can review this.

morremeyer:
@RealCLanger Rebased against master for review, and ran ruff & mksync.

I cannot explain why mksync fails in the pipeline, on my machine it does not show any changes.

RealCLanger:
I just spent some more time looking at this and testing. So, while I think it is useful to export the transactions, I think it's mostly a developer tool. And logically, the implementation of obtaining the transaction data does not really fit well into transactions.py because this currently rather is the implementation of the transaction export.
Maybe the whole feature should rather be an option of dl_docs? That it would spit out the data on the fly, just as all_events.json etc.?

morremeyer:
I personally don't think it is mostly a developer tool, since having the raw data is useful to feed into other tools.

However, I also don't care too much about it, so if you prefer that I move it to dl_docs, I can do that.

morremeyer:
@RealCLanger Do you have an opinion on my last comment?

RealCLanger:
> @RealCLanger Do you have an opinion on my last comment?

Yes, I agree, it is useful to have. In fact, I have a change near ready where I change/improve the usage of export_transactions that enables dumping of raw data as well. Unfortunately now I have to look at event parsing first. But I hope that I can soon present a PR that integrates this functionality.

morremeyer:
Great, thank you for the feedback! 

In this case, I'll just wait for the PR you have planned, let me know if you want any feedback on that.

RealCLanger:
With the latest changes, we now have a flag called `--dump-raw-data` for `dl_docs` or `export_transactions`. I believe that is what you need resp. proposed with your PR. If so, I guess this PR can be closed?

morremeyer:
@RealCLanger Thanks for the update! The formats are not 100% identical, but that's something I can easily fix in the downstream processing.

<details><summary>If anyone is interested, details here</summary>
The output with this PR is a list of objects, whereas dl_docs outputs a map of objects, where the key is the transaction ID.

In both cases, the transaction ID is available in the `id` property of the objects themselves.

When using `dl_docs`, the source is also added as a property: `"source": "timelineTransaction"`, since dl_docs has multiple sources.
</details>

Thanks for the implementation, and for keeping up all the work here!

RealCLanger:
> @RealCLanger Thanks for the update! The formats are not 100% identical, but that's something I can easily fix in the downstream processing.
> 
> If anyone is interested, details here
> The output with this PR is a list of objects, whereas dl_docs outputs a map of objects, where the key is the transaction ID.
> In both cases, the transaction ID is available in the `id` property of the objects themselves.
> 
> When using `dl_docs`, the source is also added as a property: `"source": "timelineTransaction"`, since dl_docs has multiple sources.
> 
> Thanks for the implementation, and for keeping up all the work here!

OK, great. Regarding the feedback: You mean the content of timeline_activities.json and timeline_transactions.json? I think we could adapt that, too, to add a list only. Eventually, in all_events.json, you will have a list with all these events plus data from timelineDetail.

morremeyer:
I only looked at timeline_transactions.json.

I think it would make sense to have that as a list, since having the ID as key and as a property is redundant, but it’s not that important either.

RealCLanger:
> I only looked at timeline_transactions.json.
> 
> I think it would make sense to have that as a list, since having the ID as key and as a property is redundant, but it’s not that important either.

Done: 85af79dfeacc786150bd7c9a45f24be2f58639a9

morremeyer:
Thank you!

PR #54, opened 2024-02-12
Add option to save portfolio to csv

tilidev:
- add portfolio_to_csv method to Portfolio class
- extend argument parsing with path option for saving csv
- also add help string to -o option
- portfolio_to_csv executes after get() method when path for saving csv is provided

**Core idea:**
With this option, when providing the `-o` or `--output` option (and a filepath) to the portfolio command, the key information of the portfolio is exported to a csv file. This eases the use of net value and ISIN etc. for further analysis, since csv format is easier to use when needing further analysis than the formatted string output in the CLI.


joshua-dessol:
Works great, any chance you could also add a function to export the current cash value into the CSV?

PR #65, opened 2024-05-24
Fixing issue 64

Katzmann1983:
Original fix failed if position is not tradeable, e.g. for dividends.

### DIFF
--- a/pytr/portfolio.py
+++ b/pytr/portfolio.py
@@ -38,28 +38,9 @@ class Portfolio:
 
             await self.tr.unsubscribe(subscription_id)
 
-        # Populate netValue for each ISIN
-        positions = self.portfolio['positions']
-        subscriptions = {}
-        for pos in sorted(positions, key=lambda x: x['netSize'], reverse=True):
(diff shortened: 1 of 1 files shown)


Timon0512:
Great Fix! Thank you for keeping this project alive! 
Is there a way to export the portfolio as json or csv? 

I also would love a function that loops through all isins in your portfolio and saves the details.

Timon0512:
My ultimate goal is to build a streamlit app on top of pytr. Similar to Portfolio Performance. 

As far as I understand the portfolio option prints the portfolio to the terminal. I was already able to build a funcion that saves the current portfolio as a csv. So I am fine with that. I also wrote a function that loops through all isins in your portfolio and creates seperate files with news, dividends, ratings etc. 

The only thing what I am missing right now is a complete transaction list with all buys, sells, dividends etc. I guess I can find that in the events_with_docs.json file. 
With my beginner knowlege this is quite tricky as the most informations are very nested inside that file. But I guess I will find a way. 

Thank you very much for your work :)

PR #68, opened 2024-05-24
Fix error in details command

yu-re-ka:


Katzmann1983:
I assume, the intended code was:
if not hasattr(self, 'neonNews'):

Katzmann1983:
Adding the change within https://github.com/pytr-org/pytr/pull/75

PR #95, opened 2024-07-10
Hide PIN input

DonkeeeyKong:
This hides the PIN while it is entered using `getpass()`.

I have only tested this with a Linux terminal, Windows Powershell and Windows cmd.exe. It works there. I don't know what else people use to run this though.

An alternative could be https://github.com/asweigart/pwinput. That is also closer to the way the Trade Republic app hides the PIN but it requires using an extra package.

### DIFF
--- a/pytr/account.py
+++ b/pytr/account.py
@@ -2,6 +2,7 @@ import json
 import sys
 from pygments import highlight, lexers, formatters
 import time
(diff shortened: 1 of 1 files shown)


Katzmann1983:
Love it.

DonkeeeyKong:
Thank you!

PR #107, opened 2024-07-31
timeline: retrieve next timeline, only if in the response is at least one new entry

blanke:
On my account, I get the message "Received duplicate event..." within an endless loop (all the time with the same 30 "duplicate events), so is it possible to set the variable added_last_event only, if there are no duplicates?


Katzmann1983 (approved):

Katzmann1983:
Thanks for the bug fix

Issue #116, opened 2024-09-02
Missing interest payments


felixeichler:
**Description of the bug**
In `all_transactions.csv`, the interest payments of the last 3 months is missing for me. Earlier interest payments are reflected correctly.

**To Reproduce**

1. Run `dl_docs` command
2. Inspect `all_transactions.csv`

**Expected behavior**
All interest payments should be added, or atleast the last 12 months or so. in any case not older ones and no newer ones. The sorting of all_transactions.csv is not by date, maybe this needs to be investigated to find the root cause?

**Error log**
n/a

**Environment**

- OS: macOs
- pytr version: 0.2.2

**Additional context**
Maybe this is a problem with my specific trade republic account. In the app I see the interest payments normally. 🤷‍♂️

Katzmann1983:
@felixeichler  Can you please check for todays release?

felixeichler:
@Katzmann1983 i tried updating, but it seems like no new version is in pip nor git?
<img width="615" alt="Screenshot 2024-09-04 at 09 28 55" src="https://github.com/user-attachments/assets/f4c5b77a-d041-4a62-97c4-704689003bfa">

i re-ran `dl_docs` and no change to the listed interest payments.
in my case, there's 17 interest payments in the `account_transactions.csv` from 2023-03 until 2024-06, but the ones after are missing (despite the account still received interest every month since then)...

Katzmann1983:
Sorry, screwed up the release process. please try again

felixeichler:
did try again with v0.2.4, but the results are the same

daniel-w00:
I tried with the current master branch. I get all interest payment pdfs. However, since September new pdfs are no longer in the folder `/Zinsen/Abrechnung` instead in folder `/Abrechnung`. Probabaly TR changed something on interest with the new IBAN because you now get the interests partly as dividends of money market fonds

Katzmann1983:
Sorry, I dont have the banking of Traderepublic. Can you maybe post the corresponding entry from the all_events.json?

felixeichler:
Ah yes! That's a very good pointer @daniel-w00 - because Indeed Trade Republic changed the data format of their interest payouts:
I checked my `all_events.json` and realized:
Since July 2024 they use 
```
"eventType": "INTEREST_PAYOUT",
```

and before that it was
```
"eventType": "INTEREST_PAYOUT_CREATED",
```

The other fields remain unchanged, except for changes in the details.sections incl document filepaths

[...]

PR #122, opened 2024-09-26
Refactor parsing and formatting

pinzutu:
Refactors the parsing of all_events.json and the formatting into all_transactions.csv to capture all pp events generated by tr events, especially taxes, fees, and savebacks.

In detail:
~~1. Removes card_failed_transaction type due to it not being relevant for PP~~
2. Adds INCOMING_TRANSFER_DELEGATION and events from #116 (INTEREST_PAYOUT, card_successful_oct, TAX_REFUND, benefits_saveback_execution) to event.py
3. Introduces enums and dataclasses in event.py to a. structure and facilitate type matching b. differentiate between pp_types and types that require further csv formatting
4. Introduces tax and fee parsers in event.py
5. Introduces event_formatter.py to generate multiple csv lines from certain events (saveback, ~~taxed income, transactions with fees~~). Fees and taxes are added as csv columns to avoid stand-alone transactions.
6. Adds a cli option argument to dl_docs and export_transactions to chronologically sort the exported csv transactions
Additionally:
- Changes DIVIDENDS to DIVIDEND in .po messages to respect PP's nomenclature

The all_events.json available to me parses into an accurate all_transactions.csv; however, I highly encourage everyone to test the pr on their own event logs.
I did not observe breaking changes when testing with the docker image of python 3.8

### DIFF
--- a/pytr/dl.py
+++ b/pytr/dl.py
@@ -20,6 +20,7 @@ class DL:
         history_file='pytr_history',
         max_workers=8,
         universal_filepath=False,
+        sort_export=False
     ):
         '''
         tr: api object
@@ -33,6 +34,7 @@ class DL:
         self.filename_fmt = filename_fmt
         self.since_timestamp = since_timestamp
         self.universal_filepath = universal_filepath
+        self.sort_export = sort_export
 
         self.session = FuturesSession(max_workers=max_workers, session=self.tr._websession)
         self.futures = []
--- a/pytr/event.py
+++ b/pytr/event.py
@@ -204,8 +204,6 @@ class Event:
         Returns:
             Tuple[Optional[float]]: [Tax]
         """
-        tax = None
-        return_values = {}
         # tax keywords
         tax_keys = {"Steuer", "Steuern"}
         # Gather all section dicts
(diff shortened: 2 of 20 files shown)


Katzmann1983 (approved):
Awesome changes

pinzutu:
Apologies for the small post-review change: I had the late realization that PP transactions have embedded taxes and fees fields. I therefore added taxes and fees columns to the csv file, which voids the need to generate stand-alone tax and fee events and trims event_formatter.py in the process.

Katzmann1983:
Sorry, I was first too euphoric and then tested it a little bit more. Before your last commit, taxes and fees were positive and thus gave a wrong total sum. Also I realized, that card_failed_transaction are not handled at all. These only need to be added, if and only if the transaction is of type card_failed_transaction and of status= "executed".

pinzutu:
Notwithstanding PP's indifference to the sign of fees and taxes, sticking to the negative sign convention for both makes more sense indeed.
I also wrongly believed card_failed_transaction events to always be failed transactions; these now conditionally map to REMOVALs again.

Issue #127, opened 2024-10-14
Key Error while parsing Shares and Fees.


BjBe82:
**Description of the bug**
I am running download docs and the import crashes in "event.py" after the latest changes to event parsing.

**To Reproduce**
"fees_dicts" is empty, which causes the "return_vals["fees"]) to fail. This is the part of the event send by TR. This was a fraction of a Share, which was sold separately without fees.
 
  ```
          {
                "title": "Transaktion",
                "data": [
                    {
                        "title": "Anteile",
                        "detail": {
                            "text": "0,281538",
                            "trend": null,
                            "action": null,
                            "type": "text"
                        },
                        "style": "plain"
                    },
                    {
                        "title": "Aktienkurs",
                        "detail": {
                            "text": "79,96\u00a0\u20ac",
                            "trend": null,
                            "action": null,
                            "type": "text"
                        },
                        "style": "plain"
                    },
                    {
                        "title": "Gesamt",
                        "detail": {
                            "text": "+ 22,51\u00a0\u20ac",
                            "trend": null,
                            "action": null,
                            "type": "text"
                        },
                        "style": "highlighted"
                    }
                ],
                "action": null,
                "type": "table"
            },
```

**Expected behavior**
I think the "return_vals["fees"]" call should be replaced with "return_vals["fees"]" with "return_vals.get("fees", 0)"

**Error log**

```
Exception has occurred: KeyError
'fees'
  File "E:\Trade Republic\pytr_fork\pytr\pytr\event.py", line 210, in _parse_shares_and_fees
    return return_vals["shares"], return_vals["fees"]
                                  ~~~~~~~~~~~^^^^^^^^
  File "E:\Trade Republic\pytr_fork\pytr\pytr\event.py", line 145, in _parse_type_dependent_params
    shares, fees = cls._parse_shares_and_fees(event_dict)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\Trade Republic\pytr_fork\pytr\pytr\event.py", line 106, in from_dict
    fees, isin, note, shares, taxes = cls._parse_type_dependent_params(
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\Trade Republic\pytr_fork\pytr\pytr\transactions.py", line 47, in <lambda>
    events = map(lambda x: Event.from_dict(x), timeline)
```

**Environment**

- OS: Windows 
- Local: 'de'
[...]

PR #128, opened 2024-10-14
fixed fees if not present

BjBe82:
see Issue: https://github.com/pytr-org/pytr/issues/127


pinzutu (changes requested):
Good find! For the sake of consistently formatting feeless transactions with an empty csv fees field instead of sometimes using "0.0", it is preferable for the default get value to be None.

PR #135, opened 2024-10-27
Ignore canceled deposits

RealCLanger:
It can happen that deposits get canceled. E.g. when you send money from another person's bank account which Trade Republic will reject. In that case, the deposit is canceled and no effective transaction should be created.

### DIFF
--- a/pytr/event.py
+++ b/pytr/event.py
@@ -122,6 +122,8 @@ class Event:
                 if event_dict.get("status", "").lower() == "executed"
                 else None
             )
+        elif event_type == PPEventType.DEPOSIT and event_dict.get("status", "").lower() == "canceled":
+            event_type = None
         return event_type
 
     @classmethod
--- a/pytr/event.py
+++ b/pytr/event.py
@@ -9,7 +9,6 @@ from typing import Any, Dict, Optional, Tuple, Union
 class ConditionalEventType(Enum):
     """Events that conditionally map to None or one/multiple PPEventType events"""
 
-    FAILED_CARD_TRANSACTION = auto()
     SAVEBACK = auto()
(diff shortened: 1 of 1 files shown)


pinzutu (commented):
From my logs, it seems that all events of status "CANCELED" lead to no transaction. Maybe we could simplify the method to something like this:
```suggestion
    @staticmethod
    def _parse_type(event_dict: Dict[Any, Any]) -> Optional[EventType]:
        event_type: Optional[EventType] = tr_event_type_mapping.get(
            event_dict.get("eventType", ""), None
        )
        if event_dict.get("status", "") == "CANCELED":
            event_type = None
        return event_type
```
and map "card_failed_transaction" to PPEventType.REMOVAL directly while removing ConditionalEventType.FAILED_CARD_TRANSACTION.

RealCLanger:
> From my logs, it seems that all events of status "CANCELED" lead to no transaction. Maybe we could simplify the method to something like this:
> 
> ```
>     @staticmethod
>     def _parse_type(event_dict: Dict[Any, Any]) -> Optional[EventType]:
>         event_type: Optional[EventType] = tr_event_type_mapping.get(
>             event_dict.get("eventType", ""), None
>         )
>         if event_dict.get("status", "") == "CANCELED":
>             event_type = None
>         return event_type
> ```
> 
> and map "card_failed_transaction" to PPEventType.REMOVAL directly while removing ConditionalEventType.FAILED_CARD_TRANSACTION.

I think it's a good suggestion, assuming that all events with status "CANCELED" are no-events and "card_failed_transaction" with any other status than "CANCELED" are viable deposits. I implemented it and will test it in my scenarios.

pinzutu (approved):

PR #56, opened 2024-03-14
No session resumption on different phone number

MartinScharrer:
Added code to compare given phone number with the one in the credential file and disable websession resumption if so.
For this to work well I also added a command line option to select if credentials should be saved or not.
As a bonus a fix for corrupt cookie files got added.

This pull-request fixes https://github.com/marzzzello/pytr/issues/38.

PR #144, opened 2024-11-21
Improve handling of multiple credentials

### LINKED ISSUE #38: Session is not invalidated when a new login with different number occurs
**Description of the bug**
When working with multiple accounts even if a different phone number and pin are supplied the previous session gets resumed.

**To Reproduce**
Login using `pytr login` using an existing key file. Then `pytr login -n <somenumber> -p <somepin>`. Might as well save a new key file. Now `pytr dl_docs` will use the first account not the one which was logged in last (given the session is still valid).

**Expected behavior**
When logging in with a number other than the one the active session is linked to, I expect a the session to be invalidated and a new session to start.

**Error log**

**Environment**

- OS: MacOS
- pytr version: 0.1.5

RealCLanger:
Hi,

I'm taking another attempt to fix #38 (Session is not invalidated when a new login with different number occurs) which hits me as well. There has already be an effort in #56 but it seems that this went stale.

My idea is to store the cookies in a file that is named after the phone number which is the id of a trade republic account. So there can be no conflicts in cookies if the tool is used with multiple accounts.

I also remove the prompt that asks whether one would like to save the credentials and replace it with a command line option `--store_credentials` which defaults to true. I find the prompt rather annoying, especially for scripted calls of pytr.

Comments/Suggestions?

Cheers
Christoph


pytr/account.py
```diff
@@ -52,19 +51,13 @@ def login(phone_no=None, pin=None, web=True):
...
                 f.writelines([phone_no + "\n", pin + "\n"])
 
             log.info(f"Saved credentials in {CREDENTIALS_FILE}")
 
-        else:
-            save_cookies = False
-            log.info("Credentials not saved")
-
-    tr = TradeRepublicApi(phone_no=phone_no, pin=pin, save_cookies=save_cookies)
+    tr = TradeRepublicApi(phone_no=phone_no, pin=pin, save_cookies=True)
```
pinzutu, line 60: Same critique here, the default behavior should not be to leave clear file cookies on your system indefinitely.
```suggestion
        else:
            save_cookies = False
            log.info("Credentials not saved")

    tr = TradeRepublicApi(phone_no=phone_no, pin=pin, save_cookies=save_cookies)
```

pinzutu (changes requested):
That aside, being able to store multiple credentials seems like a useful feature to have👍

RealCLanger:
> That aside, being able to store multiple credentials seems like a useful feature to have👍

OK, I think that is a good suggestion. I changed the default to not store credentials. So one would need to explicitly turn that behavior on.

Issue #143, opened 2024-11-20
websockets 14.1 produces error


tjalb:
**Description of the bug**
When running `pytr dl_docs ./` it ends with `BaseEventLoop.create_connection() got an unexpected keyword argument 'extra_headers'`. This is caused by a change in the "websockets" package version 14.

**To Reproduce**

1. Go to the TradeRepublic document folder
2. Ran the command `pytr dl_docs ./`
3. Enter the code from the TR-App'

**Expected behavior**
pytr downloads the documents.

**Error log**

```
File "/opt/homebrew/bin/pytr", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/pytr/main.py", line 266, in main
    asyncio.get_event_loop().run_until_complete(dl.dl_loop())
  File "/opt/homebrew/Cellar/python@3.11/3.11.10/Frameworks/Python.framework/Versions/3.11/lib/python3.11/asyncio/base_events.py", line 654, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/pytr/dl.py", line 68, in dl_loop
    await self.tl.get_next_timeline_transactions()
  File "/opt/homebrew/lib/python3.11/site-packages/pytr/timeline.py", line 30, in get_next_timeline_transactions
    await self.tr.timeline_transactions()
  File "/opt/homebrew/lib/python3.11/site-packages/pytr/api.py", line 506, in timeline_transactions
    return await self.subscribe({"type": "timelineTransactions", "after": after})
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/pytr/api.py", line 322, in subscribe
    ws = await self._get_ws()
         ^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/pytr/api.py", line 301, in _get_ws
    self._ws = await websockets.connect(
               ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/websockets/asyncio/client.py", line 442, in __await_impl__
    self.connection = await self.create_connection()
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #145, opened 2024-11-29
fixed extra_headers and ws.open

nitro424:
- extra_headers was renamed to additional_headers
- using close_code to check if _ws is open


Mohamed-elg (approved):

hjacobs:
This will fix #143

kacnje:
When will this commit be released?

pytr/api.py
```diff
@@ -263,7 +263,7 @@ def _web_request(self, url_path, payload=None, method="GET"):
         return self._websession.request(method=method, url=f"{self._host}{url_path}", data=payload)
 
     async def _get_ws(self):
-        if self._ws and self._ws.open:
+        if self._ws and self._ws.close_code == None:
```
NiklasRosenstein, line 266:
```suggestion
        if self._ws and self._ws.close_code is None:
```

Satisfy linter

dharmanibc:
Hi.., the same error persists and the solution doesn't work. Got the conflicting requirements by google genai installed: ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
google-genai 1.20.0 requires websockets<15.1.0,>=13.0.0, but you have websockets 10.1 which is incompatible.

PR #150, opened 2024-12-09
Set correct project url

timo-keller:
The url to the homepage on PyPI points to a repository on Gitlab which apparently does not exist.


nakhli (approved):
LGTM

PR #157, opened 2025-01-13
Increase netSize precision to 6 decimal when exporting portfolio in CSV

angt:
When exporting the portfolio to CSV, we lose precision by limiting to only 3 decimal places for some stocks.


pytr/portfolio.py
```diff
@@ -94,7 +94,7 @@ def portfolio_to_csv(self, output_path):
         csv_lines = []
         for pos in sorted(positions, key=lambda x: x["netSize"], reverse=True):
             csv_lines.append(
-                f"{pos['name']};{pos['instrumentId']};{float(pos['netSize']):>10.3f};{float(pos['averageBuyIn']):.2f};{float(pos['netValue']):.2f}"
+                f"{pos['name']};{pos['instrumentId']};{float(pos['netSize']):>10.6f};{float(pos['averageBuyIn']):.2f};{float(pos['netValue']):.2f}"
```
NiklasRosenstein, line 97: It looks like the field is a decimal string already, I don't see why we're converting it to a float and back to a string in the first place.

```json
{
  "instrumentId": "KYG9830T1067",
  "netSize": "85.0",
  "averageBuyIn": "5.5508",
  "averageAccruedInterest": "0.0",
  "name": "Xiaomi",
  "exchangeIds": [
    "LSX",
    "TDG",
    "TIB"
  ],
  "netValue": 452.88000000000005
}
```

Issue #134, opened 2024-10-26
Minor issue: no newline after last line in portfolio CSV export


marcusfey:
This is really a small problem/question.

Is it intended that there is no \n after the portfolio content [here](https://github.com/pytr-org/pytr/blob/3280e88ad5c4d5824a797e9db5345e2549cf008b/pytr/portfolio.py#L92)?

RealCLanger:
Probably not. Is the \n important to you? I guess you can open a PR that adds it then... 😄

PR #158, opened 2025-01-16
add newline to last line of portfolio_to_csv output

marcusfey:
Fix for #134


pytr/portfolio.py
```diff
@@ -97,9 +97,10 @@ def portfolio_to_csv(self, output_path):
                 f"{pos['name']};{pos['instrumentId']};{float(pos['netSize']):>10.3f};{float(pos['averageBuyIn']):.2f};{float(pos['netValue']):.2f}"
             )
 
+
```
NiklasRosenstein, line 100:
```suggestion
```

PR #160, opened 2025-01-21
Pin `websockets<14`

NiklasRosenstein:
Addresses https://github.com/pytr-org/pytr/issues/143 until someone comes along and updates the code to be compatible with websockets 14.x


marzzzello:
@NiklasRosenstein is this still necessary when https://github.com/pytr-org/pytr/pull/145 is applied?

Issue #151, opened 2024-12-16
Error while using ptr dl_docs


HiidanX:
**Description of the bug**
I have been getting an error for two weeks running pytr dl_docs.

**To Reproduce**

2. Ran the command `pytr dl_docs`

**Expected behavior**
Download of all pdf files from Trade Republic

**Error log**

/.local/pipx/venvs/pytr/lib/python3.12/site-packages/pytr/timeline.py", line 130, in process_timelineDetail
    event = self.timeline_events[response["id"]]
KeyError: '05fcd862-75f4-4271-8236-60ebf207b2f2'

**Environment**

- OS: Mac
- pytr version: 0.3.1

**Additional context**
This used to work for me without a problem. Now there seems to be a problem. Anyone else noticed similar behaviour?
Root Cause: Christmas present

HiidanX:
I have narrowed down the problem. When including the parameter last_days parameter it seems to work. There seems to be problem when exporting the christmas present (Weihnachtsgeschenk). Is there a way to fix this?

peterandree:
doesn't work for me even with --last days

peterandree:
> doesn't work for me even with --last days

have to correct myself. When I triangulate last days so, that the christmas gift is not in the timespan, it seems to work. Good to know, but somewhat useless since I need to download older values too

HiidanX:
Yes, a bugfix is needed to somehow correct the christmas present. Has anyone solved the problem already?

HiidanX:
It seems to the same error as the issue #148

otmezger:
In my case, it seems to be breaking with a `stock gift`. I had 2 of them in December.

NiklasRosenstein:
Closing as duplicate of #148 (and that issue was addressed).

Issue #148, opened 2024-12-03
KeyError while reading timeline event for gift feature


pleserwolf:
**Description of the bug**
While downloading new files using pytr dl_docs, I now get a KeyError after all timeline events have been requested. It looks like response["id"] returns a UUID that is then not found in the timeline. The referenced UUID changes with every call of pytr dl_docs.

I have tried both running the command in my documents folder that is populated with a history of pytr documents as well as in a clean new folder. The only file that gets created in the new folder is an empty pytr_history.

Printing out response, it seems like the error is related to a transaction by the new "gift" feature.

```
{'id': 'REDACTED UUID', 'sections': [{'title': 'Du hast REDACTED\xa0€ geschenkt', 'data': {'icon': 'logos/contacts-M-Orange/v2', 'timestamp': 'REDACTED', 'status': 'executed'}, 'type': 'header'}, {'title': 'Übersicht', 'data': [{'title': 'Status', 'detail': {'text': 'Angenommen', 'functionalStyle': 'EXECUTED', 'type': 'status'}, 'style': 'plain'}, {'title': 'Asset', 'detail': {'text': 'REDACTED ETF NAME', 'type': 'text'}, 'style': 'plain'}, {'title': 'Lieferdatum', 'detail': {'text': '02 Dec', 'type': 'text'}, 'style': 'plain'}], 'type': 'table'}, {'title': 'Status', 'steps': [{'leading': {'avatar': {'status': 'completed', 'type': 'bullet'}, 'connection': {'order': 'first'}}, 'content': {'title': 'Geschenk geliefert', 'timestamp': 'REDACTED'}}, {'leading': {'avatar': {'status': 'completed', 'type': 'bullet'}, 'connection': {'order': 'last'}}, 'content': {'title': 'Geschenk angenommen', 'timestamp': 'REDACTED'}}], 'type': 'steps'}, {'title': 'Transaktion', 'data': [{'title': 'Geschenkbetrag', 'detail': {'text': 'REDACTED \xa0€', 'type': 'text'}, 'style': 'plain'}, {'title': 'Gebühr', 'detail': {
[...]

PR #161, opened 2025-01-21
Ignore unsupported Timeline events and log them

NiklasRosenstein:
The Christmas Gift event was causing errors, see

* https://github.com/pytr-org/pytr/issues/148
* https://github.com/pytr-org/pytr/issues/151

There might be a sensible way to handle this event, but before that I think it makes sense to make it so the `dl_docs` command doesn't outright crash when there's an unsupported event in the timeline.

Inspired largely by @zonefuenf (https://github.com/pytr-org/pytr/issues/148#issuecomment-2532089877)

### DIFF
--- a/pytr/dl.py
+++ b/pytr/dl.py
@@ -8,7 +8,7 @@ from pathvalidate import sanitize_filepath
 
 from pytr.utils import preview, get_logger
 from pytr.api import TradeRepublicError
-from pytr.timeline import Timeline
+from pytr.timeline import Timeline, UnsupportedEventError
 
 
 class DL:
@@ -82,7 +82,10 @@ class DL:
             elif subscription.get("type", "") == "timelineActivityLog":
                 await self.tl.get_next_timeline_activity_log(response)
             elif subscription.get("type", "") == "timelineDetailV2":
-                await self.tl.process_timelineDetail(response, self)
+                try:
+                    self.tl.process_timelineDetail(response, self)
+                except UnsupportedEventError:
+                    self.log.warning("Ignoring unsupported event %s", response)
             else:
                 self.log.warning(
                     f"unmatched subscription of type '{subscription['type']}':\n{preview(response)}"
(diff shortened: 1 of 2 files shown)


pytr/timeline.py
```diff
@@ -119,15 +123,18 @@ async def _get_timeline_details(self):
...
+    def process_timelineDetail(self, response, dl):
         """
         process timeline details response
         download any associated docs
         create other_events.json, events_with_documents.json and account_transactions.csv
         """
 
+        event = self.timeline_events.get(response["id"], None)
+        if event is None:
+            raise UnsupportedEventError(response["id"])
```
JustinTrvz, line 135: Add the same check for `get_next_timeline_activity_log()` and `get_next_timeline_transactions()` to be consistent.

pytr/dl.py
```diff
@@ -82,7 +82,10 @@ async def dl_loop(self):
             elif subscription.get("type", "") == "timelineActivityLog":
                 await self.tl.get_next_timeline_activity_log(response)
             elif subscription.get("type", "") == "timelineDetailV2":
-                await self.tl.process_timelineDetail(response, self)
+                try:
+                    self.tl.process_timelineDetail(response, self)
+                except UnsupportedEventError:
+                    self.log.warning("Ignoring unsupported event %s", response)
```
JustinTrvz, line 88: What if an exception occures while calling `get_next_timeline_transactions()` or `get_next_timeline_activity_log()`?
I would suggest to surround the whole if-else cascade with a try and catch.

JustinTrvz (changes requested):
Approach is too focused on `process_timelineDetail()`. Please keep the other event processing functions in mind.

pytr/dl.py
NiklasRosenstein: I don't want to catch arbitrary errors as that includes the category of "unknown unknowns". Not understanding a particular event type in the code is a "known unknown" that we can relatively safely work around with (here, by ignoring the event). Other types of errors might indicate a whole different kind of mistake in the logic (or change in the TR protocol) that might be disfigured beyond recognition by catching and simply ignoring it (even with added logs).

PR #169, opened 2025-01-30
Add Ruff formatting & linting

NiklasRosenstein:
This PR adds [Ruff] as a development dependency and uses it in CI to check for lints and to enforce formatting.

[Ruff]: https://astral.sh/ruff

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -5,7 +5,7 @@ on:
   - workflow_dispatch
 
 jobs:
-  check:
+  checks:
     name: Checks
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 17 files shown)


NiklasRosenstein:
Seems you gave me permissions to merge without waiting for your review, @marzzzello. I'll wait anyway to be sure. 🙂

RealCLanger:
Cool, I think this looks so much nicer that using "black".

However, in the readme, we have a reference to black:
https://github.com/pytr-org/pytr?tab=readme-ov-file#linting-and-code-formatting

Can you mention ruff there and, if it conflicts (which I believe it does), maybe drop black there?

RealCLanger (approved):

NiklasRosenstein:
Indeed, thanks for the notice. I've updated the README now

PR #172, opened 2025-01-31
Fix missing date in filename

### LINKED ISSUE #137: MINOR: Some files contain leading blanks, where than nextcloud complains, and hopefully small feature request
**Description of the bug**
Some filenames contain occassionally leading blanks, which leads nextcloud to "complain", and it suggest a rename.

**To Reproduce**
Just download. In my case it happened with "Transaktionsbestätigung", for an US stock. It happens rarely, I could not find a pattern. So I suggest to "sanitize" the filename "somehow". My beloved suggestion is below as a feature request.

**Environment**

- OS: Mac OS
- pytr version: 0.3.0

**Additional context**
I have _a lot_ downloads, as I trade quite often. And I could not find a pattern when it happens. It occurs in a very rare case. If I would like to have a wish, I would like to suggest a more "fixed?" "static" filename like
YYYY-MM-DD-HH-MM-SS-type-tr-transaction-number-description.pdf

where 
YYYY-MM-... is the transaction date/time as accurate as you can get it. It would help finding problems.
"type"  is "Abrechnung", "Basisinformationsblatt", ...
"tr-transaction-number" looks like 047713ed-84c1-49a1-8331-e6c10e764ad5 which is added if I trade a stock twice or more a day it seems. It is missing the first time, but from than on it is added.
"description" is e.g. "E.L.F. Beauty".

remove all leading blanks, and maybe trailing as well as you cannot "see" them really in the file listings.
replace the test of the blanks with an underscore.

All in all as always: Many thanks for this great piece of software. I can't live without it!

thanks
juergen

Gabscap:
Problem: Downloaded documents are sometimes missing the date in the file name. This is caused by `doc['detail']` not containing a date string.

Fix: If `doc['detail']` does not contain a date, it uses the time from `event['timestamp']`.

This also fixes #137

### DIFF
--- a/pytr/dl.py
+++ b/pytr/dl.py
@@ -93,7 +93,7 @@ class DL:
                     f"unmatched subscription of type '{subscription['type']}':\n{preview(response)}"
                 )
 
-    def dl_doc(self, doc, titleText, subtitleText, subfolder=None):
+    def dl_doc(self, doc, titleText, subtitleText, subfolder=None, timestamp=None):
         """
         send asynchronous request, append future with filepath to self.futures
         """
@@ -105,8 +105,12 @@ class DL:
             date = doc["detail"]
             iso_date = "-".join(date.split(".")[::-1])
         except KeyError:
-            date = ""
-            iso_date = ""
+            if timestamp:
+                date = timestamp.strftime("%d.%m.%Y")
+                iso_date = timestamp.strftime("%Y-%m-%d")
+            else:
+                date = ""
+                iso_date = ""
         doc_id = doc["id"]
 
         # extract time from subtitleText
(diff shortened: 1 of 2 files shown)


NiklasRosenstein:
Thanks @Gabscap , can confirm this is working.

```
-out/Transaktionsbestätigung/ Transaktionsbestätigung - Niklas Rosenstein (4b5e90e6-3979-4734-b187-af98fb7cdfb1).pdf
+out/Transaktionsbestätigung/2024-12-04 Transaktionsbestätigung - Niklas Rosenstein (4b5e90e6-3979-4734-b187-af98fb7cdfb1).pdf
```

PR #177, opened 2025-02-11
Run CI on Pull Requests

NiklasRosenstein:
Update the `on` configuration of the workflow:

- Trigger on pushes to the main branch (`master`)
- Trigger on any pull requests
- Remove `workflow_dispatch` trigger (it appears to be unused)

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -1,10 +1,11 @@
 name: Build and publish Python distributions to PyPI
 
 on:
-  - push:
-      branches: ["master"]
-  - pull_request
-  - workflow_dispatch
+  push:
+    branches: [ 'master' ]
+    tags: [ '*' ]
+  pull_request:
+    branches: [ '*' ]
 
 jobs:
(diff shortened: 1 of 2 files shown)


RealCLanger:
@NiklasRosenstein Wouldn't the workflow_dispatch be useful when a pypi release should be updated?

NiklasRosenstein:
> @NiklasRosenstein Wouldn't the workflow_dispatch be useful when a pypi release should be updated?

What do you mean by that?

As it is today (and also was before this PR), the release was published from the Job on the `push` event. e.g. here for `v0.2.4`:

image

https://github.com/pytr-org/pytr/actions/runs/10698846420

RealCLanger:
> As it is today (and also was before this PR), the release was published from the Job on the `push` event. e.g. here for `v0.2.4`:

Well yes, but couldn't it be one would want to update a release, or a release build fails? But, ok, yes, if one (re-)pushes a tag, the releasing should be triggered. So I guess it's fine to remove the workflow_dispatch trigger.

PR #181, opened 2025-02-11
Upgrade PyPI publish version from v1.9 to v1 (i.e. using the latest in the major version 1)

NiklasRosenstein:
This is necessary because unfortunately the version that was pinned did not support Python package metadata version 2.4, which is the version that `uv build` with Hatchling produced.

```
Metadata-Version: 2.4
Name: pytr
Version: 0.4.0
Summary: Use TradeRepublic in terminal
...
```

This made the release job for 0.4.0 fail, unfortunately.


NiklasRosenstein:
@marzzzello Could you update the `v0.4.0` tag and force push it? Alternatively, we can make a `v0.4.1` tag instead.

PR #186, opened 2025-02-12
Map `card_tr_refund` transaction to `DEPOSIT`

### LINKED ISSUE #164: VISA Chargeback transactions not included in `account_transactions.csv`
VISA Chargeback transactions (i.e. when VISA is refunding money paid to a dealer that didn't provide the owed services, e.g. due to insolvency) aren't included in `pytr`'s `account_transactions.csv`.

Such event looks like the following (note `"eventType": "card_tr_refund"`):

```json
  {
    "id": "00000000-0000-0000-0000-000000000000",
    "timestamp": "2024-08-20T00:00:00.000+0000",
    "title": "Erstattung",
    "icon": "logos/bank_traderepublic/v2",
    "badge": null,
    "subtitle": null,
    "amount": {
      "currency": "EUR",
      "value": 123.45,
      "fractionDigits": 2
    },
    "subAmount": null,
    "status": "EXECUTED",
    "action": {
      "type": "timelineDetail",
      "payload": "00000000-0000-0000-0000-000000000000"
    },
    "eventType": "card_tr_refund",
    "cashAccountNumber": null,
    "hidden": false,
    "deleted": false,
    "source": "timelineTransaction",
    "details": {
      "id": "00000000-0000-0000-0000-000000000000",
      "sections": [
        {
          "title": "Du hast 123,45 € erhalten",
          "data": {
            "icon": "logos/bank_traderepublic/v2",
            "timestamp": "2024-08-20T00:00:00.000+0000",
            "status": "executed"
          },
          "type": "header"
        },
        {
          "title": "Übersicht",
          "data": [
            {
              "title": "Status",
              "detail": {
                "text": "Fertig",
                "functionalStyle": "EXECUTED",
                "type": "status"
              },
              "style": "plain"
            },
            {
              "title": "Zahlungsart",
              "detail": {
                "text": "Rückerstattung",
                "type": "text"
              },
              "style": "plain"
            }
          ],
          "type": "table"
        },
        {
          "title": "Transaktion",
          "data": [
            {
              "title": "FTI",
              "detail": {
                "title": "FTI",
                "timestamp": "2024-05-20T00:00:00.000+0000",
                "amount": "123,45 €",
                "icon": "logos/merchant-e681d92a-c850-414e-8be4-1d6bc9e44adc/v2",
                "status": "executed",
                "action": {
                  "payload": "00000000-0000-0000-0000-000000000000",
                  "type": "timelineDetail"
                },
                "type": "embeddedTimelineItem"
              },
              "style": "plain"
            }
          ],
          "type": "table"
        }
      ]
    },
    "has_docs": false
  },
```

Thank you for sharing your work with us, keep up the great work, thank you! ❤ 

**Environment**

- OS: Docker Container `docker.io/python:3.13.1-bookworm`
- pytr version: pytr 0.3.1

NiklasRosenstein:
Potentially fixes #164

### DIFF
--- a/pytr/event.py
+++ b/pytr/event.py
@@ -21,7 +21,7 @@ class PPEventType(Enum):
     DEPOSIT = "DEPOSIT"
     DIVIDEND = "DIVIDEND"
     FEES = "FEES"  # Currently not mapped to
-    FEES_REFUND = "FEES_REFUND"  # Currently not mapped to
+    FEES_REFUND = "FEES_REFUND"
     INTEREST = "INTEREST"
     INTEREST_CHARGE = "INTEREST_CHARGE"  # Currently not mapped to
     REMOVAL = "REMOVAL"
@@ -73,6 +73,8 @@ tr_event_type_mapping = {
     "SAVINGS_PLAN_INVOICE_CREATED": ConditionalEventType.TRADE_INVOICE,
     "benefits_spare_change_execution": ConditionalEventType.TRADE_INVOICE,
     "TRADE_INVOICE": ConditionalEventType.TRADE_INVOICE,
+    # Fees refund
+    "card_tr_refund": PPEventType.FEES_REFUND,
 }
 
 
--- a/pytr/event.py
+++ b/pytr/event.py
@@ -46,6 +46,7 @@ tr_event_type_mapping = {
     "PAYMENT_INBOUND_SEPA_DIRECT_DEBIT": PPEventType.DEPOSIT,
     "card_refund": PPEventType.DEPOSIT,
     "card_successful_oct": PPEventType.DEPOSIT,
+    "card_tr_refund": PPEventType.DEPOSIT,
     # Dividends
     "CREDIT": PPEventType.DIVIDEND,
     "ssp_corporate_action_invoice_cash": PPEventType.DIVIDEND,
@@ -72,8 +73,6 @@ tr_event_type_mapping = {
     "SAVINGS_PLAN_INVOICE_CREATED": ConditionalEventType.TRADE_INVOICE,
     "benefits_spare_change_execution": ConditionalEventType.TRADE_INVOICE,
     "TRADE_INVOICE": ConditionalEventType.TRADE_INVOICE,
-    # Fees refund
-    "card_tr_refund": PPEventType.DEPOSIT,
 }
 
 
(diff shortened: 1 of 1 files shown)


PhrozenByte (changes requested):
`pytr` now takes the transaction into account, thanks :+1:

However, I noticed that you mapped `card_tr_refund` to `PPEventType.FEES_REFUND`. I don't think that this is correct and probably just a misunderstanding: This transaction isn't about fees, but about a normal transaction with a 3rd-party merchant. The only difference to `card_refund` is that `card_refund` is initiated by the merchant, whereas `card_tr_refund` is initiated by VISA resp. Trade Republic as part of the VISA Consumer Dispute ("VISA Chargeback") process. See https://www.visa.co.uk/chargeback.html (German / Deutsch: https://www.visa.de/visa-everywhere/blog/bdp/2024/06/07/kartenzahlung-schuetzt-verbraucherinnen-1717763716229.html) for more details about VISA Chargeback. Mastercard has a similar process, but isn't relevant for Trade Republic. Thus, `card_tr_refund` should probably be mapped to `PPEventType.DEPOSIT` instead.

NiklasRosenstein:
Thanks for the explanation @PhrozenByte, I think mapping it to `DEPOSIT` makes sense.

pytr/event.py
```diff
@@ -25,7 +25,7 @@ class PPEventType(EventType):
     DEPOSIT = "DEPOSIT"
     DIVIDEND = "DIVIDEND"
     FEES = "FEES"  # Currently not mapped to
-    FEES_REFUND = "FEES_REFUND"  # Currently not mapped to
+    FEES_REFUND = "FEES_REFUND"
```
RealCLanger, line 28: Now this change is not correct any longer because FEES_REFUND is still not mapped to.

RealCLanger (changes requested):

pytr/event.py
NiklasRosenstein: You're right, thanks

NiklasRosenstein:
```suggestion
    FEES_REFUND = "FEES_REFUND"  # Currently not mapped to
```

RealCLanger (approved):

pytr/event.py
```diff
@@ -72,6 +72,8 @@ class PPEventType(EventType):
     "SAVINGS_PLAN_INVOICE_CREATED": ConditionalEventType.TRADE_INVOICE,
     "benefits_spare_change_execution": ConditionalEventType.TRADE_INVOICE,
     "TRADE_INVOICE": ConditionalEventType.TRADE_INVOICE,
+    # Fees refund
+    "card_tr_refund": PPEventType.DEPOSIT,
```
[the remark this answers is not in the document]
NiklasRosenstein: Indeed 🙈

PR #187, opened 2025-02-13
Introduce basic Mypy type checking

NiklasRosenstein:
This PR adds Mypy to the dev dependencies, CI and makes minimal changes to the code to get Mypy to pass.


RealCLanger (approved):
Looks good to me and seems a sensible thing to do.

PR #188, opened 2025-02-13
Clean up some comments and minor code improvement

RealCLanger:
minor cleanups

### DIFF
--- a/pytr/portfolio.py
+++ b/pytr/portfolio.py
@@ -44,7 +44,7 @@ class Portfolio:
 
             if subscription["type"] == "instrument":
                 await self.tr.unsubscribe(subscription_id)
-                pos = subscriptions.pop(subscription_id, None)
+                pos = subscriptions.pop(subscription_id)
                 pos["name"] = response["shortName"]
                 pos["exchangeIds"] = response["exchangeIds"]
             else:
--- a/pytr/portfolio.py
+++ b/pytr/portfolio.py
@@ -65,7 +65,7 @@ class Portfolio:
 
             if subscription["type"] == "ticker":
                 await self.tr.unsubscribe(subscription_id)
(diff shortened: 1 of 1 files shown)


NiklasRosenstein (approved):
~LGTM 👍~

See below

pytr/portfolio.py
```diff
@@ -51,8 +44,7 @@ async def portfolio_loop(self):
 
             if subscription["type"] == "instrument":
                 await self.tr.unsubscribe(subscription_id)
-                pos = subscriptions[subscription_id]
-                subscriptions.pop(subscription_id, None)
+                pos = subscriptions.pop(subscription_id, None)
```
NiklasRosenstein, line 47: Actually, at least typing wise, this leaves a chance for `pos` to be `None`, and changes the current semantic of the code (`NoneType object is not subscriptable`), as opposed to raising a `KeyError`.

```suggestion
                pos = subscriptions.pop(subscription_id)
```

pytr/portfolio.py
```diff
@@ -73,8 +65,7 @@ async def portfolio_loop(self):
 
             if subscription["type"] == "ticker":
                 await self.tr.unsubscribe(subscription_id)
-                pos = subscriptions[subscription_id]
-                subscriptions.pop(subscription_id, None)
+                pos = subscriptions.pop(subscription_id, None)
```
NiklasRosenstein, line 68:
```suggestion
                pos = subscriptions.pop(subscription_id)
```

See above

NiklasRosenstein (changes requested):

pytr/portfolio.py
RealCLanger: Good point. Even better.

PR #189, opened 2025-02-14
Refactor event exporter and add JSON output format

NiklasRosenstein:
This PR refactors the `EventCsvFormatter` class, renames it to `TransactionExporter` and moves it into `pytr.transactions`.

- Use `csv.DictWriter` instead of homegrown CSV formatting
  - note: With the default dialect, which at the current state in this PR is not configurable, the CSV delimiter changes from `;` (semicolon) to `,` (comma)
- Add option to disable localization of values in transactions (`--no-decimal-localization`). Allows that numbers can be stored as actual numbers instead of strings in JSON
- Changed the default configuration to disable decimal localization and to include the timestamp in the date field.
  - These defaults are used by `Timeline.finish_timeline_details()` that writes out `account_transactions.csv`
  - However the CLI default values for `export_transactions` are still the same, so the output will be the same as before except for the CSV delimiter
  - (I'd like to move towards adopting these defaults but I'm not sure if this will cause people to run into issues because they might be unaware of the change when they upgrade and whatever people load the CSV into might not support not-localized numbers and commas.. ?)


RealCLanger:
Overall, looks like a good refactoring/enhancement. Please give me some time to play with it.

pytr/main.py
```diff
@@ -184,25 +188,46 @@ def formatter(prog):
         "input",
         help="Input path to JSON (use other_events.json from dl_docs)",
         metavar="INPUT",
-        type=Path,
+        type=argparse.FileType("r"),
+    )
+    parser_export_transactions.add_argument(
+        "output",
+        help="Output path of CSV file",
```
RealCLanger, line 195: Maybe rather "Output path of export file"? It could as well be json now.

pytr/transactions.py
```diff
@@ -1,60 +1,174 @@
...
+    lang: str = "en"
+    """ The language for the CSV header / JSON keys. """
+
+    date_with_time: bool = True
+    """ Include the timestamp in ISO8601 format in the date field. """
+
+    decimal_localization: bool = False
+    """ Whether to localize the decimal format. If enabled, decimal fields will be string values. """
+
+    csv_delimiter: str = ";"
```
RealCLanger, line 67: This is not used atm because you hard code ";" down below.

RealCLanger (commented):
I looked at the change a bit more in detail. Nice refactoring, especially that it does not read in a file again. I made two minor comments. Have not yet tried this in practice but planning to do so, soon.

pytr/transactions.py
NiklasRosenstein: Good catch!

pytr/transactions.py
```diff
@@ -1,60 +1,174 @@
...
+    A helper class to convert Trade Republic events each to one or more line items that are a simplified representation
+    useful for a importing for example into a portfolio manager.
     """
-    log = get_logger(__name__)
-    if lang == "auto":
-        locale = getdefaultlocale()[0]
-        if locale is None:
-            lang = "en"
+
+    lang: str = "en"
```
RealCLanger, line 58: Maybe we want to use the default "auto" here, too?

pytr/main.py
```diff
@@ -184,25 +188,46 @@ def formatter(prog):
...
+        choices=["auto", *sorted(SUPPORTED_LANGUAGES)],
         default="auto",
     )
     parser_export_transactions.add_argument(
-        "--date-isoformat",
-        help="Format the date column in ISO8601 including the time.",
-        action="store_true",
+        "--date-with-time",
+        action=argparse.BooleanOptionalAction,
+        default=True,
```
RealCLanger, line 211: This changes the default behavior. Not sure we want that? Maybe should default to False here?

RealCLanger (approved):
OK, now I played around with it a bit in practice, too. Looks really good, just two remarks with regards to default values which I leave open to you to decide. 👍

pytr/transactions.py
```diff
@@ -1,60 +1,174 @@
...
+        format: Literal["json", "csv"] = "csv",
+    ) -> None:
+        self._log.info("Exporting transactions ...")
+        if sort:
+            events = sorted(events, key=lambda ev: ev.date)
+
+        transactions = (txn for event in events for txn in self.from_event(event))
+
+        if format == "csv":
+            writer = csv.DictWriter(fp, fieldnames=self.fields(), delimiter=self.csv_delimiter)
```
RealCLanger, line 166: Another thing: On Windows I have to add `lineterminator="\n"` here, otherwise I get a strange file format with empty lines in between.

pytr/transactions.py
```diff
@@ -1,60 +1,174 @@
...
+
+        kwargs: _SimpleTransaction = {
+            "date": event.date.isoformat() if self.date_with_time else event.date.date().isoformat(),
+            "type": self._translate(event.event_type.value) if isinstance(event.event_type, PPEventType) else None,
+            "value": self._decimal_format(event.value),
+            "note": self._translate(event.note) + " - " + event.title if event.note is not None else event.title,
+            "isin": event.isin,
+            "shares": self._decimal_format(event.shares, False),
+            "fees": self._decimal_format(-event.fees) if event.fees is not None else None,
+            "taxes": self._decimal_format(event.taxes),
```
RealCLanger, line 132: This had a minus previously, so it should be `"taxes": self._decimal_format(event.taxes),`. Any reason for changing that?

pytr/transactions.py
NiklasRosenstein: I'd much prefer to require explicitly setting `lang="auto"`. The "auto" value here effectively introduces a dependency on the system locale, which is anything but obvious if the default value is `"auto"`.

Though I think this goes for code, mostly, and this need not apply to UI/UX. So how about we add a `--lang` option where it is missing that defaults to `auto`?

pytr/main.py
NiklasRosenstein: See the "additional notes" section above 🙂 

I think there is an argument for changing the default, though there could also be a middle way where we warn about default values that will change in the future and only a few versions later we could change it.. ?

pytr/transactions.py
NiklasRosenstein: What editor are you opening the file with? According to [RFC 4180 and Mime standards](https://en.wikipedia.org/wiki/Comma-separated_values#RFC_4180_and_MIME_standards), CSV files are expected to use CRLF line terminators. Thus, that is what the [`csv.Dialect.lineterminator`](https://docs.python.org/3.6/library/csv.html#csv.Dialect.lineterminator) defaults to.

pytr/transactions.py
NiklasRosenstein: That's a good catch, thank you.

NiklasRosenstein, line 132:
```suggestion
            "taxes": self._decimal_format(-event.taxes),
```

NiklasRosenstein:
```suggestion
            "taxes": self._decimal_format(-event.taxes),
```

pytr/transactions.py
RealCLanger: OK, sounds good.

pytr/main.py
RealCLanger: OK, yes, makes sense. I leave it to you whether we want to delay the switch of the default or do it now. I lean, however, towards changing the default now because I don't see a reason for holding it back, given that there's a switch to turn on the old behavior.

pytr/transactions.py
RealCLanger: I discussed this with ChatGPT and learned that the reason is that "If you open the file in regular text mode ("w"), Python itself converts \n to \r\n on Windows.". So we effectively result in \r\r\n which is one \r too much. I have a suggestion what to do to improve things.

pytr/transactions.py
```diff
@@ -1,60 +1,174 @@
...
+        format: Literal["json", "csv"] = "csv",
+    ) -> None:
+        self._log.info("Exporting transactions ...")
+        if sort:
+            events = sorted(events, key=lambda ev: ev.date)
+
+        transactions = (txn for event in events for txn in self.from_event(event))
+
+        if format == "csv":
+            writer = csv.DictWriter(fp, fieldnames=self.fields(), delimiter=self.csv_delimiter)
```
RealCLanger, line 166:
```suggestion
            lineterminator = "\n" if platform.system() == "Windows" else "\r\n"
            writer = csv.DictWriter(fp, fieldnames=self.fields(), delimiter=";", lineterminator=lineterminator)
```

pytr/main.py
```diff
@@ -184,25 +188,46 @@ def formatter(prog):
...
         help="Input path to JSON (use other_events.json from dl_docs)",
         metavar="INPUT",
-        type=Path,
+        type=argparse.FileType("r"),
+    )
+    parser_export_transactions.add_argument(
+        "output",
+        help="Output file path",
+        metavar="OUTPUT",
+        type=argparse.FileType("w"),
```
RealCLanger, line 197:
```suggestion
        type=argparse.FileType("w", encoding="utf8"),
```

RealCLanger: We should keep the old behavior to set the encoding to utf-8.

RealCLanger (approved):
Made some last comments. :)

pytr/transactions.py
```diff
@@ -1,60 +1,174 @@
...
+
+        kwargs: _SimpleTransaction = {
+            "date": event.date.isoformat() if self.date_with_time else event.date.date().isoformat(),
+            "type": self._translate(event.event_type.value) if isinstance(event.event_type, PPEventType) else None,
+            "value": self._decimal_format(event.value),
+            "note": self._translate(event.note) + " - " + event.title if event.note is not None else event.title,
+            "isin": event.isin,
+            "shares": self._decimal_format(event.shares, False),
+            "fees": self._decimal_format(-event.fees) if event.fees is not None else None,
+            "taxes": self._decimal_format(-event.taxes),
```
RealCLanger, line 132:
```suggestion
            "taxes": self._decimal_format(-event.taxes) if event.taxes is not None else None,
```

You need to handle None here.

pytr/timeline.py
```diff
@@ -198,10 +200,12 @@ def finish_timeline_details(self, dl):
                 indent=2,
             )
 
-        export_transactions(
-            dl.output_path / "all_events.json",
-            dl.output_path / "account_transactions.csv",
-            sort=dl.sort_export,
-        )
+        with (dl.output_path / "account_transactions.csv").open("w") as f:
```
RealCLanger, line 203:
```suggestion
        with (dl.output_path / "account_transactions.csv").open("w", encoding="utf8") as f:
```
Should be utf8 here, too.

pytr/transactions.py
NiklasRosenstein: That analysis seems correct. I also found this SO post about it: https://stackoverflow.com/questions/3191528/csv-in-python-adding-an-extra-carriage-return-on-windows

pytr/transactions.py
NiklasRosenstein: I've also found this SO question about the same issue: https://stackoverflow.com/questions/3191528/csv-in-python-adding-an-extra-carriage-return-on-windows

The answer given here seems a more appropriate solution, considering it is also what is recommended in the documentation of the `csv` module: Opening the file with `newline=''` to disable universal newline.

RealCLanger: yes, as written above, you could open the file with `newline=''`. However the output file for exporting transactions is opened through `type=argparse.FileType("w", encoding="utf8"),` in main.py and this does not support the newline option. But I guess you could refactor that.

NiklasRosenstein: 🤦  Sorry for overlooking that you already noted that. You're right, `argparse.FileType` does not accept the argument. People in [Issue24739](https://bugs.python.org/issue24739) encourages you to create your own `FileType`-like custom class that accepts the `newline` argument 🤔

RealCLanger:
I added some fixes and tweaks. I implemented my windows line ending fix, although it maybe could be done differently. I also add the new options for export_transactions to dl_docs since this implicitly exports transactions in the end as well.

I will merge this now to bring the tool forward. I guess we can still change/fix things through subsequent PRs.

One thing to note is that default behavior will be changed by this. We will now have timestamps with time in account_transactions.csv, which could be turned off through `--no-date-with-time` and no decimal localization takes place which could be turned on with `--decimal-localization`

PR #192, opened 2025-02-14
Ignore uv.lock

RealCLanger:
This file gets created on my Windows. Should be ignored for git.


RealCLanger:
If it makes sense to add the file, rather than ignoring it, please feel free to do so. I'm not too knowledgeable in that area 😉

Issue #141, opened 2024-11-10
[BUG] Export Account Transactions has wrong tax


ralfescher:
Hi,

I am a beginner in Python programming and managed to get this (great) program running under a virtual Python 3.9 on my new Mac mini.
 
While exporting the transactions, I noticed that the taxes were incorrectly displayed.

Output: 2024-11-01;Interest;187.16;Zinsen;;;;**-6,704**
The real value is: 187,16€ and **67,04**€ tax (instead of 6704€).

Additionally, I have the problem of importing .csv files into Excel on my Mac (German locale) 
(comma/period and UTF-8 -> Mac-Roman). 
Would it make sense to offer a specific option for this?

_Additional question: Does it make sense to get this app running under Python 3.12 as well? (As mentioned, I’m new to Python)._

Thanks
Ralf

pinzutu:
Hi, you can change pytr's csv locale using the export_transactions --lang argument. Excel should also be able to [parse semicolumn separated csvs](https://www.ablebits.com/office-addins-blog/change-excel-csv-delimiter/). 

Furthermore, could you share the (redacted) json entry of the event with the erroneous tax (all_events.json)?

msdn65:
In account_transactions.csv I also get wrong or malformatted tax:
2024-11-01;INTEREST;146.51;Zinsen;;;;-3,137

snippet from all_events.json:
        {
          "title": "Transaktion",
          "data": [
            {
              "title": "Angesammelt",
              "style": "plain",
              "detail": {
                "text": "€177.88",
                "type": "text"
              }
            },
            {
              "title": "Steuern",
              "style": "plain",
              "detail": {
                "text": "€31.37",
                "type": "text"
              }
            },
            {
              "title": "Gesamt",
              "style": "plain",
              "detail": {
                "text": "€146.51",
                "type": "text"
              }
            }
          ],
          "type": "table"
        },

ralfescher:
I have the same situation: the values in the JSON look correct (as with @msdn65)

SegFaulty:
> Hi, you can change pytr's csv locale using the export_transactions --lang argument. 

chnaging the language does not change the outcome, only the format ... tax is wrong 
```
transaction-de.csv:2024-11-01;INTEREST;224,34;Zinsen;;;;-8.035
transaction-en.csv:2024-11-01;INTEREST;224.34;Zinsen;;;;-8,035
``` 
80.35 would by correct

BUT I digged deeper:

[...]

PR #193, opened 2025-02-14
Improve number parsing and handle a few more events

RealCLanger:
This tries to address a few things.

First of all I added two additional event types that occured in my depot lately: `trading_savingsplan_executed` which seems to be something like `SAVINGS_PLAN_EXECUTED` or `SAVINGS_PLAN_INVOICE_CREATED`. And additionally `TRADE_CORRECTED` which is like `TRADE_INVOICE`.

In order to identify new events in the future I thought it's a good idea to log a warning when unknown events occur. I added code for that.

Furthermore I try to address issue with locales and number parsing, which still seem to exist as per the recent discussion in #141 (https://github.com/pytr-org/pytr/issues/141#issuecomment-2654090421). My idea is to give the german local a general preference before english. With that, at least in my account, going through the whole history, I could not spot regressions. However, if anybody observes something, please let me know.

Generally, most numbers will be parsed correctly because it can be something like `1.234,56` - german which will fail to be parsed as english and `1,234.56` which will fail with the german locale. An issue would now be a situation where you have a full english localized in the range of 1000 - 999999 that could be written like `1,000` and would now be parsed as 1.0 instead of 1000.0.

But as the number format in TR data seems to be wildly mixed it is really hard to predict. Maybe we could change the locale preference for some specific types of events + titles - feedback welcome.


NiklasRosenstein (approved):

pytr/event.py
```diff
@@ -247,13 +280,12 @@ def _parse_float_from_detail(elem_dict: Dict[str, Any]) -> Optional[float]:
...
+            return None
         parsed_val = re.sub(r"[^\,\.\d-]", "", unparsed_val)
 
-        # Try the locale that will fail more likely first.
-        if "." not in parsed_val:
-            locales = ("en", "de")
-        else:
-            locales = ("de", "en")
+        # Prefer german locale.
+        locales = ("de", "en")
```
MedAzizKhayati, line 288: This will introduce a new issue, sometimes the TR API delivers numbers in American format, this will result in errors such as:
10.002 getting parsed as 10002 instead of 10.002 when parsed by german first.

RealCLanger: ok, can you name specific events where that would happen? As I said, at least in my TR account I did not encounter a problem. But it might be just a coincidence.

pytr/event.py
```diff
@@ -110,9 +138,14 @@ def from_dict(cls, event_dict: Dict[Any, Any]):
...
-        if event_dict.get("status", "").lower() == "canceled":
-            event_type = None
+        eventTypeStr = event_dict.get("eventType", "")
+        event_type: Optional[EventType] = tr_event_type_mapping.get(eventTypeStr, None)
+        if event_type is not None:
+            if event_dict.get("status", "").lower() == "canceled":
+                event_type = None
+        else:
+            if eventTypeStr not in events_known_ignored:
+                log.warning(f"Ignoring event {eventTypeStr}")
```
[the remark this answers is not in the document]
RealCLanger: I made the log a bit more explicit and I also introduce an option `--dump-debug-data-to-file` which would print out more data that could then be provided to developers. WDYT?

pytr/event.py
RealCLanger: I put some more effort into this and checked all the events in my timeline where `en` should be preferred before `de` when parsing numbers. I also added logs/warnings and an option to dump more data that could help developers.

pytr/event.py
```diff
@@ -110,9 +139,16 @@ def from_dict(cls, event_dict: Dict[Any, Any]):
...
+        eventTypeStr = event_dict.get("eventType", "")
+        event_type: Optional[EventType] = tr_event_type_mapping.get(eventTypeStr, None)
+        if event_type is not None:
+            if event_dict.get("status", "").lower() == "canceled":
+                event_type = None
+        else:
+            if eventTypeStr not in events_known_ignored:
+                log.warning(f"Ignoring unknown event {eventTypeStr}")
+                if dump_enabled():
+                    dump(f"Unknown event {eventTypeStr}: {pprint.pformat(event_dict, indent=4)}")
```
NiklasRosenstein, line 151: I think this could be handled with `logging` just as well, don't you think? And formatting the `event_dict` as JSON would make it easier to handle, for example to include as sample data in tests.

```suggestion
                self._log.debug("Unknown event %r: %s", eventTypeStr, json.dumps(event_dict, indent=4)))
```

Where the `--dump-debug-data-to-file` option is handled, we can add a [`FileHandler`](https://docs.python.org/3/library/logging.handlers.html#filehandler) to the root logger.

```py
handler = logging.FileHandler(args.dump_debug_data_to_file)
handler.setLevel(logging.DEBUG)
handler.setFormatter(logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s'))
logging.root.addHandler(handler)
```

And maybe shorten the option name to `--debug-logfile`?

Alternatively, we could forego the entire option and add a `--debug` or `--verbose` option to set the log level to `DEBUG` such that these messages show up in the console and people can grab it from there; or alternatively have a `--debug --logfile myfile.log` as a combination to achieve the same effect.

RealCLanger: OK, I'll mull about this. The reason for the `if dump_enabled():` construct is that pretty printing (or json formatting) the data always is quite expensive.

RealCLanger:
I think this PR is in quite a good shape now. I'll merge it tomorrow, if there's no further comments.

PR #195, opened 2025-02-20
Handle bonds calculation in portfolio

aleksandrzhiliaev:
Hi folks, just found that bonds calculation gives wrong numbers in portfolio list that also gives wrong amount of Depot sum in the end, that PR fixes netValue by assuming that we are dealing with bonds in name.


NiklasRosenstein:
Hi @aleksandrzhiliaev , thanks for your contribution!

Would you be able to provide an example event data for a bonds transaction?

I'm also thinking if this change would be better suited in another place, or needed _also_ in another place, because there's a few more places that use the same API and don't handle bonds special (e.g. exporting to CSV).

aleksandrzhiliaev:
Unfortunately, I currently don't have any bond transactions. I want to accurately calculate my portfolio since bonds are not included in the calculations right now. Where could I incorporate bond logic in a more general way?

RealCLanger:
Thanks for your contribution. This makes sense and I could verify it with a bond in my portfolio. Merged.

PR #215, opened 2025-05-10
fix: report corrected fractional amounts for savings plans and report price

kaihowl:
The API response from TR will contain floating point errors for saving plans when fractional shares are bought. The transaction report as well as the PDF report do not contain this error.

For now paper over this deficiency by reporting the total amount from the transaction details instead in those cases.

Example from my actual output:
```
    "amount": {
      "currency": "EUR",
      "value": -499.999874,
      "fractionDigits": 2
    },
```

Additionally, this also reports the price of the bought instrument. This serves as additional help to verify the correct amounts when importing into for example gnucash.

### DIFF
--- a/pytr/event.py
+++ b/pytr/event.py
@@ -132,6 +132,7 @@ class Event:
     note: Optional[str]
     shares: Optional[float]
     taxes: Optional[float]
+    price: Optional[float]
     value: Optional[float]
 
     @classmethod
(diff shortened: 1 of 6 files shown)


kaihowl:
Unfortunately, there are also cases where we have fractional cents in the price for the instruments on the official PDF document. Those are not reported in the (raw) output of the Trade Republic API. I don't know how to fix that problem at the moment.

RealCLanger:
Thanks for bringing up this PR. As I see right now, you are adding price and transaction value fields, which is a good addition, I guess. But I don't see changes that relate to the handling of fraction digits (apart from a small change in test input). Am I missing something?

PR #218, opened 2025-05-20
More pythonic API by using `asyncio.Queue`

nagy:
This is my attempt to make the API more pythonic.

For each subscription on the WebSocket we are now creating a `asyncio.Queue`.

https://docs.python.org/3/library/asyncio-queue.html

This way, we do not have to count the number of received responses, but can let `asyncio.gather` handle all that calculation for us as you can see in `details.py`. The usage of `asyncio.Queue` allows further simplification in other files, which are not in this PR. `recv2` creates a loop that is feeding these queues with newly arrived messages and is therefore a `asyncio.Task` that needs to be started at the beginning. This is currently done in `details_loop` but I would rather collect that at a more centralized place.

https://docs.python.org/3/library/asyncio-task.html#asyncio.Task

The intention is not to introduce these new `2`-suffixed function, but to show how the other functions could look like.
If the overall idea looks good to you, then I can change all the other functions as well to follow this scheme and undraft this PR.

### DIFF
--- a/pytr/api.py
+++ b/pytr/api.py
@@ -66,6 +66,7 @@ class TradeRepublicApi:
     _subscription_id_counter = 1
     _previous_responses: Dict[str, str] = {}
     subscriptions: Dict[str, Dict[str, Any]] = {}
+    subscriptions_futures: Dict[str, asyncio.Queue] = {}
 
     _credentials_file = CREDENTIALS_FILE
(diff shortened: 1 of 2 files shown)


ExploracuriousAlex:
That looks nice. This could also make the Timeline module cleaner and better readable.

PR #219, opened 2025-05-23
Fix missing shares and fees in account_transactions.csv

nagy:
The export of the `account_transactions.csv` file is currently missing the shares and fees amounts. This PR fixes that but the tests are currently still broken. If you can confirm, that those numbers are also missing for you, then I can adapt the unit test for this as well.


RealCLanger:
Thanks for the report/suggestion. I'm seeing this as well and also have a fix prepared. Will post it later on.

nagy:
Superseded by #223 .

PR #222, opened 2025-05-30
Fix: Saveback and some Sparplan transactions lack shares information

derterz:
I noticed that in the CSV export a lot of transactions had no "Stück" information. It seems this was mainly related to "Savebacks" and "Sparplan" types. At least for me the CSV was fine with this commit.

### DIFF
--- a/pytr/event.py
+++ b/pytr/event.py
@@ -265,7 +265,6 @@ class Event:
 
             break
 
-
         return shares, fees
 
     @classmethod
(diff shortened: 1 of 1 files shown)


RealCLanger:
Thanks for proposing this fix suggestion for a current issue. However, I have just created PR #223 that also addresses this along with other changes and fixes. Unfortunately I didn't have time to do this earlier such that your work could have been avoided, sorry for that.

In any case, would you mind checking and testing #223? Thanks!

derterz:
I just tested the solution from #223 and indeed the result is almost identical. The only difference is that with my solution one transaction fee is missing.

Issue #212, opened 2025-05-01
OSError: [Errno 63] File name too long


bluefish007:
I get the following error: OSError: [Errno 63] File name too long:

The problem is with old documents from 2020, where the doc_id is part of the filename (at least for one document in my case) and is written again in parentheses after it in line 136 in dl.py:
`filename_with_doc_id = filename + “f” ({doc_id})”`

This makes the filename too long!

Possible fix for line 136 in dl.py:
```
if filename.endswith(doc_id):
    filename_without_id = filename[:-len(doc_id)].rstrip()
    filename_with_doc_id = filename_without_id + f" ({doc_id})"
else:
    filename_with_doc_id = filename + f" ({doc_id})"
```

RealCLanger:
Wouldn't it be enough to just use:

```
if filename.endswith(doc_id):
    filename_with_doc_id = filename
else:
    filename_with_doc_id = filename + f" ({doc_id})"
```
?

bluefish007:
Yes, that would also be possible, but then you would have “filename doc_id.pdf” and “filename (doc_id).pdf”

I think it would be nicer if it was always the same, i.e. always in brackets, therefore my suggestion.

RealCLanger:
OK, understood. I added the code to PR #223.

RealCLanger:
Since the changes for the PR were merged and a new version released (v0.4.3) I'm closing this issue as resolved. Please let me know if there's still a problem with that fix.

PR #223, opened 2025-05-31
Fix parsing of new TR API event style and a few minor things

RealCLanger:
Trade Republic has modified the payload for some events. This needs some fixes in event parsing.

With this change I also add some event types that went unrecognized so far and I incorporate a suggestion about file names from #212.

### DIFF
--- a/pytr/event.py
+++ b/pytr/event.py
@@ -271,12 +271,14 @@ class Event:
 
         transaction_dict = next(filter(lambda x: x.get("title") in ["Transaktion"], sections), None)
         if transaction_dict:
+            # old style event
             dump_dict["maintitle"] = transaction_dict["title"]
             data = transaction_dict.get("data", [{}])
             shares_dict = next(filter(lambda x: x["title"] in ["Aktien", "Anteile"], data), None)
             fees_dict = next(filter(lambda x: x["title"] == "Gebühr", data), None)
             taxes_dict = next(filter(lambda x: x["title"] in ["Steuer", "Steuern"], data), None)
         else:
+            # new style event
             uebersicht_dict = next(filter(lambda x: x.get("title") in ["Übersicht"], sections), None)
             if uebersicht_dict:
                 for item in uebersicht_dict.get("data", []):


pytr/dl.py
```diff
@@ -133,7 +133,10 @@ def dl_doc(self, doc, titleText, subfolder, doc_date):
             id=doc_id,
         )
 
-        filename_with_doc_id = filename + f" ({doc_id})"
+        # In case, the filename already ends with the doc id, we remove it to avoid a duplicate id in the name
+        filename_with_doc_id = (
+            filename[: -len(doc_id)].rstrip() if filename.endswith(doc_id) else filename
+        ) + f" ({doc_id})"
```
nagy, line 139: Is this the same as `filename.removesuffix(doc_id)+ f" ({doc_id})"`?

pytr/event.py
```diff
@@ -224,46 +246,115 @@ def _parse_isin(event_dict: Dict[Any, Any]) -> str:
...
+            None,
+        )
         dump_dict = {"eventType": event_dict["eventType"], "id": event_dict["id"]}
 
         sections = event_dict.get("details", {}).get("sections", [{}])
-        transaction_dicts = filter(lambda x: x.get("title") in ["Transaktion"], sections)
-        for transaction_dict in transaction_dicts:
+
+        transaction_dict = next(filter(lambda x: x.get("title") in ["Transaktion"], sections), None)
+        if transaction_dict:
```
nagy, line 274: So you want to keep supporting the old format? Is that because past entries are still being returned in this old format or just for other compatibility? I may have missed it but for me it did not seem to return entries in the old format.

nagy (commented):
Code looks good so far. I have yet to test it.

derterz:
Tested it and indeed it works nicely for me. Thanks a lot!

pytr/event.py
```diff
@@ -92,26 +94,48 @@ def get_event_logger():
...
+    "INPAYMENTS_SEPA_MANDATE_CREATED",
+    "INSTRUCTION_CORPORATE_ACTION",
+    "GENERAL_MEETING",
+    "GESH_CORPORATE_ACTION",
+    "MATURITY",
     "ORDER_CANCELED",
     "ORDER_CREATED",
     "ORDER_EXPIRED",
     "ORDER_REJECTED",
+    "PAYMENT-SERVICE-IN-PAYMENT-DIRECT-DEBIT",
```
decontamin4t0R, line 119: This caused some direct transfers to be missing from the generated file...

pytr/dl.py
RealCLanger: Yes, good catch.

pytr/event.py
RealCLanger: Yes, I think the old format could still be around. Better be safe here.

pytr/event.py
RealCLanger: Hm, ok, removed for now. However, what kind of event is this? It is not mapped to anything through tr_event_type_mapping...

pytr/event.py
nagy: Could we add a comment here so that future readers know that this is the branch for the older variant?

RealCLanger: I tested it with a fresh run on all my data and find both, old and new style events. So I added comments but definitely need to keep the old style parsing around.

PR #225, opened 2025-06-19
Fix dividends on knock-outs

nagy:
On (knockout-)options, before this it would assign the ISIN of the underlying instrument instead of the ISIN of the option itself and the amount of options was also missing.

If you want I can also craft a testcase for this.

### DIFF
--- a/pytr/event.py
+++ b/pytr/event.py
@@ -252,6 +252,7 @@ class Event:
             if event_type is PPEventType.DIVIDEND:
                 isin = cls._parse_isin(event_dict)
                 taxes = cls._parse_taxes(event_dict)
+                shares, _, _, _ = cls._parse_shares_fees_taxes_and_value(event_dict)
 
             elif event_type is PPEventType.INTEREST:
                 taxes = cls._parse_taxes(event_dict)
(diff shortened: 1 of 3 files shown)


RealCLanger:
I guess this makes sense overall.

It would be good to have a test case with an added example event, also to have some data at hand for developers that don't have such events at hand in their own account.

nagy:
I have added a test for this.

pytr/event.py
```diff
@@ -207,6 +207,7 @@ def _parse_type_dependent_params(
         if event_type is PPEventType.DIVIDEND:
             isin = cls._parse_isin(event_dict)
             taxes = cls._parse_taxes(event_dict)
+            shares, _, _ = cls._parse_shares_fees_and_taxes(event_dict)
```
RealCLanger, line 210: remove the _parse_taxes call above and get taxes from the new call to _parse_shares_fees_and_taxes

pytr/event.py
```diff
@@ -317,6 +323,20 @@ def _parse_shares_fees_and_taxes(
                     else:
                         continue
 
+            # The case for dividend payments on knock-outs
+            for shares_dict in filter(lambda x: x.get("title") in ["Aktien entfernt"], uebersicht_dict.get("data", [])):
```
RealCLanger, line 327: I think you can get rid of this section if you add something like this in line 304, when going through uebersicht_dict:

                    if title == "Aktien entfernt":
                        shares_dict = item
                        if fees_dict and taxes_dict:
                            break
                        else:
                            continue

nagy:
The tests seem to pass and it is a lot cleaner now. Thank you.

RealCLanger:
I just submitted larger changes for parsing to the mainline. Can you merge these changes into your PR, then we can finalize this.

nagy:
Done. Thank you for your work.

RealCLanger:
> Done. Thank you for your work.

Thank you, I've merged it.

PR #230, opened 2025-07-07
ci: fix python version

morremeyer:
This fixes the python version selection for the CI tests.


RealCLanger:
Good catch. Now please also remove python version 3.9 because hatch-babel does not support it, then we can merge.

morremeyer:
@RealCLanger Done!

PR #245, opened 2025-10-13
Fix typo in method name

Tubinex:
Fix typo: rename inititate_weblogin to initiate_weblogin


RealCLanger:
You also need to update account.py then

Tubinex:
You’re absolutely right, I missed that earlier! It’s fixed now.

Issue #240, opened 2025-08-21
SEPA Direct Debit Transactions (Lastschriften) missing in the account_transaction.csv


eprora:
**Description of the bug**
Hello I encountered the following problem when rechecking my transactions that SEPA Direct Debit are missing from the list of transactions.

**To Reproduce**

1. Have a SEPA Direct Debit (Lastschrift) booked as a prerequisite (e.g. electricity, internet provider)
2. Ran the command `pytr dl_docs`
3. Checked 'account_transactions.csv'

**Expected behavior**
The transaction is part of 'account_transactions.csv'.

**Error log**
The transaction is not part of 'account_transactions.csv'.

**Environment**

- OS: Windows 10
- pytr version: 0.4.3

**Additional context**
nothing

RealCLanger:
Do you see an output like: "Ignoring unknown event" ?

eprora:
Hello, sorry I was on vacation. I got two different messages in the output:

`no mapping for DIRECT_DEBIT_TRANSFER_DELEGATION`

`Ignoring unknown event DIRECT_DEBIT_TRANSFER_DELEGATION`

RealCLanger:
I think this issue should now be obsolete. Please post a comment/reopen should it still exist.

PR #248, opened 2025-10-21
Add DIRECT_DEBIT_TRANSFER_DELEGATION event type to removals

eprora:
Addresses issue #240 to support previously not support SEPA direct debit transactions.


MartinScharrer:
TR seems to have removed the eventType field from the JSON response!

RealCLanger:
I just saw that now. In theory a good change but since the eventType field was removed by trade republic we must rewrite the parsing completely, I think. I'm looking at it...

MartinScharrer:
It looks like the type must now be detected on the other fields with some heuristic method. 

Maybe a nice `match` construct would be possible.

The document type for the downloads is still there, so that could be a hint.

RealCLanger:
I think this PR is now obsolete.

Issue #253, opened 2025-11-03
pytr exits with KeyError: 'eventType'


fft1010:
**Description of the bug**
I just run pytr like this:

`pytr dl_docs ../2025-tr-pytr-docs`

It worked this morning at 10:44

**Error log**

```
14:58:14 Timeline activity log: Subscribing to #1...
14:58:15 Timeline activity log: Received #1, subscribing to #2...
14:58:15 Timeline activity log: Received #2, subscribing to #3...
14:58:15 Timeline activity log: Received #3 (last relevant).
14:58:15 Requested all timeline details (157/157).
14:58:15   1/157: BioNTech (ADR) -- Kauforder - 2025-11-03T13:35:33
Traceback (most recent call last):
  File "/Users/ry90/.cache/uv/archive-v0/vF7hz6gWjTFJRrgLbNevX/bin/pytr", line 12, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/ry90/.cache/uv/archive-v0/vF7hz6gWjTFJRrgLbNevX/lib/python3.12/site-packages/pytr/main.py", line 354, in main
    asyncio.get_event_loop().run_until_complete(dl.dl_loop())
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/Users/ry90/.cache/uv/archive-v0/vF7hz6gWjTFJRrgLbNevX/lib/python3.12/site-packages/pytr/dl.py", line 85, in dl_loop
    await self.tl.process_timelineDetail(response, self)
  File "/Users/ry90/.cache/uv/archive-v0/vF7hz6gWjTFJRrgLbNevX/lib/python3.12/site-packages/pytr/timeline.py", line 226, in process_timelineDetail
    if event["eventType"] == "timeline_legacy_migrated_events":
       ~~~~~^^^^^^^^^^^^^
KeyError: 'eventType'
```

**Environment**

- no change since last working version
- OS:  MAC OS 14.8.1 (23J30)
- pytr version: 0.4.3

RealCLanger:
This is bad. TR have removed the eventType key from their events - all parsing is based on that. Let's see if and how this can be fixed...

kugel-:
I can confirm this issue.

ProtectOurPlayground:
I can confirm it too.
This is bad :(

[...]

PR #255, opened 2025-11-06
partial workaround for missing eventTypes: new dl_raw option to at least download all data from TR and beeing able to parse them offline

ExploracuriousAlex:
Similar to dl_docs, dl_raw will save all events to the files all_events.json, events_with_documents.json, and other_events.json, and download all documents with their original name. This will allow us to take further steps to adapt event parsing, file renaming, and sorting offline, now where TR does not provide eventTypes anymore. (see #253)

I didn't want to merge this with the master branch since I also changed the async loop behavior for Python 3.14.
(Replacing `asyncio.get_event_loop().run_until_complete()` with `asyncio.run()`).
This should also work with previous Python versions, though I have not tested it.

The focus here was on making changes as simple as possible and ensuring non-breaking functionality (even though TR broke the old function). Using this branch/version you can at least download the event data and documents. And then you can take your time "offline" to parse the events.

### DIFF
--- a/pytr/dl_raw.py
+++ b/pytr/dl_raw.py
@@ -4,6 +4,7 @@ from pathlib import Path
 from requests_futures.sessions import FuturesSession  # type: ignore[import-untyped]
 
 from pytr.api import TradeRepublicError
+from pytr.post_sort import post_sort
 from pytr.timeline import Timeline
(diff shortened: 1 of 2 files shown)


ExploracuriousAlex:
No longer required.

PR #258, opened 2025-11-07
Replace event type mapping

Corentin-tin:
…e handling in transactions


RealCLanger:
Thanks for providing this PR. This goes into the same direction as my just integrated #259. However, everything is still a WIP and I'll continue to make changes in that area. But I'm really glad for any help with testing and other suggestions...

PR #261, opened 2025-11-08
added "Dividende" in addition to "Bardividende" to subtitle_event_type_mapping

ExploracuriousAlex:
subtitle_event_type_mapping supplemented and tests added/adjusted


RealCLanger (approved):
Good. One more. 😄 And also thanks for adding the test.

PR #264, opened 2025-11-10
Handle unrequested events

ExploracuriousAlex:
Instead of ignoring unrequested event responses, this change will create pseudo time activities and attach the details of such previously unrequested events.

Why this?

During the promotion in December 2024, participants could enter a prize draw by donating a share (worth €25 or more) and win up to €10,000.

Previously, documents for corresponding actions (e.g., securities settlements for these specific promotions) were ignored because there is no timeline activity for them, only "unrequested event responses" (timelineDetailV2).

With this change also these documents are now downloaded.

This change is non-breaking.

### DIFF
--- a/pytr/timeline.py
+++ b/pytr/timeline.py
@@ -235,10 +235,30 @@ class Timeline:
 
         event = self.timeline_details.get(response.get("id", "dummy"), None)
         if event is None:
-            self.log.warning(f"Ignoring unrequested event response {json.dumps(response, indent=4)}")
-            self.skipped_detail += 1
-            self.finish_if_done()
-            return
+            # Extract timestamp from response sections if available
+            sections = response.get("sections", [])
+            pseudo_timestamp = "1999-12-31T23:59:59.999+0000"
+            timestamp = pseudo_timestamp
+
+            # Search for timestamp in any section and data entry
+            for section in sections:
+                data = section.get("data", [])
+                if isinstance(data, list):
+                    for data_item in data:
+                        if isinstance(data_item, dict) and "timestamp" in data_item:
+                            timestamp = data_item["timestamp"]
+                            break
+                elif isinstance(data, dict) and "timestamp" in data:
+                    timestamp = data["timestamp"]
+                    break
+
+                if timestamp != pseudo_timestamp:
+                    break
+
+            event = {"id": response["id"], "title": "Pseudo", "subtitle": "Pseudo", "timestamp": timestamp}
+            self.timeline_activities[response["id"]] = event
+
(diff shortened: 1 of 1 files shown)


RealCLanger:
Hm, I can not see how that works. So, you add a dummy event (for timeline detail) with literally no data, except for the bare minimum of timestamp, title and subtitle and find additional documents to download? There are no new URLs in this dummy data. But maybe I'm missing something...

ExploracuriousAlex:
What you miss is that the details event with download URL (paypload) etc. is then attached to this pseudo timeline event and thus gets processed instead of beeing dropped.

RealCLanger:
> What you miss is that the details event with download URL (paypload) etc. is then attached to this pseudo timeline event and thus gets processed instead of beeing dropped.

Yes, but you now add a dummy details event with no data. I am missing the place where real new data (with URLs) is received from TR. AFAICS, there is no timeline details response.

ExploracuriousAlex:
`process_timelineDetail()` is only called because indeed a **timeline detail event** arrived. Usually there is a existing _timeline activity event_ were this is attached as detail. But in some special cases there is no _timeline activity event_. Until now such **timeline detail events** were then just skipped.

With my change, they (the **timeline detail events**) get processed and corresponding documents get downloaded.

Please see my comments directly in "Files changed". There it might get clearer. It is line 264. There the incoming **timeline detail event** (response) is added to the newly created pseudo _timeline activity event_.

With that change I get 3 more documents from last years promotion event.

RealCLanger:
> `process_timelineDetail()` is only called because indeed a **timeline detail event** arrived. Usually there is a existing _timeline activity event_ were this is attached as detail. But in some special cases there is no _timeline activity event_. Until now such **timeline detail events** were then just skipped.
> 
> With my change, they (the **timeline detail events**) get processed and corresponding documents get downloaded.
> 
> Please see my comments directly in "Files changed". There it might get clearer. It is line 264. There the incoming **timeline detail event** (response) is added to the newly created pseudo _timeline activity event_.
> 
> With that change I get 3 more documents from last years promotion event.

Ah, I see. In your case we get unexpected, additional timeline details, e.g. without having found a transaction or activity log beforehand which we then have to synthesize.

I think we then need to get the bookkeeping correct by also incrementing self.requested_detail. Otherwise I think we'll miss events, don't we?

Furthermore, maybe we can also try to find a title/subtitle for the new event from the received detail?

ExploracuriousAlex:
After taking a closer look at how details (timelineDetailV2) are matched with timeline events (timelineTransactions,  timelineActivityLog), I’m discarding this workaround in favor of a better solution.

Issue #183, opened 2025-02-11
Include Gift event in transaction


NiklasRosenstein:
End of last year, TR introduced a feature to send gifts (e.g. some amount of stock) via email to be accepted by the recipient into their TR portfolio. This event is currently being ignored due to https://github.com/pytr-org/pytr/pull/161 (instead of causing an error), but it also means the event remains entirely unprocessed.

From what I can tell, there is no document related to the particular gifting event. (There are events for accepting the Terms & Conditions related to gifting, which are separate and already handled)

However, this is still a transaction that is missing from `pytr export_transactions`. This is because the event is not saved into the `all_events.json` which is used as input to the command.

## Research notes

#161 skips the `timelineDetailV2` for which there is not corresponding timeline event.

https://github.com/pytr-org/pytr/blob/f830c55a59f5d8233a06e7328184886601583ff5/pytr/timeline.py#L128-L130

However, it appears there _is_ a corresponding timeline event, only the mapping we currently do based on the `id` does not match (at least in my case).

```py
(Pdb) [ev for k, ev in self.timeline_events.items() if "Geschenk" in str(ev)]
[
    {
        "id": "316dbb9c-c94b-44f8-b3ea-3a90d27021b5",
        "timestamp": "2024-12-14T17:55:19.707+0000",
        "title": "ETF-Geschenk",
        "icon": "logos/contacts-B-Orange/v2",
        "badge": None,
        "subtitle": "Angenommen",
        "amount": {
            "currency": "EUR",
            "value": -26.0,
            "fractionDigits": 2
        },
        "subAmount": None,
        "status": "EXECUTED",
        "action": {
            "type": "timelineDetail",
            "payload": "316dbb9c-c94b-44f8-b3ea-3a90d27021b5"
        },
        "eventType": "GIFTER_TRANSACTION",
        "cashAccountNumber": None,
        "hidden": False,
        "deleted": False,
        "source": "timelineTransaction"
    }
]
```

<details>

<summary>Corresponding `timelineDetailV2`<summary>

```json
{
    "id": "882bb1e7-8a14-42a3-b026-42a1d61d699a",
    "sections": [
        {
            "title": "Du hast 25,00\xa0€ geschenkt",
            "data": {
                "icon": "logos/contacts-B-Orange/v2",
                "timestamp": "2024-12-14T17:55:19.707+0000",
                "status": "executed"
            },
            "type": "header"
        },
        {
            "title": "Übersicht",
            "data": [
                {
[...]

PR #266, opened 2025-11-12
Improved matching of details to timeline events

ExploracuriousAlex:
Up to now, **details** (responses of the type `timelineDetailV2`) have been linked to **timeline events** (responses of the type `timelineTransactions `or `timelineActivityLog`) via their IDs, under the assumption that these match.  Here is the code location:

```
async def process_timelineDetail(self, response):
    """
    process timeline details response
    """

    event = self.timeline_details.get(response.get("id", "dummy"), None)
```

However, there are exceptions. For instance, in my case, only 96% of **details** share the same ID as their corresponding **timeline event**. The remaining ones have so far been treated as "unrequested event response". But in reality, there are no unrequested responses. The only issue lies in the matching process.

Some **details** even include a `timelineEventId `attribute. Unfortunately, matching via this attribute turns out to be even less reliable. For example, in my case, only 26% of the `timelineEventId` values actually match the ID of the corresponding **timeline event**.

Therefore, I’ve added a new matching approach based on the subscription ID. For me, this method achieves a 100% success rate. 

By default, the old evend ID-based matching method is still used. For the new matching method, there’s a new parameter `--details-matching` with the choices `sub_id` (new method) and `event_id` (previous method) available for both `dl_docs` and `export_transactions`.

### DIFF
--- a/pytr/timeline.py
+++ b/pytr/timeline.py
@@ -200,8 +200,7 @@ class Timeline:
 
             self.requested_detail += 1
             if msg is None:
-                subscription_id = await self.tr.timeline_detail_v2(event["id"])
-                event["timeline_detail_subscription_id"] = subscription_id
+                await self.tr.timeline_detail_v2(event["id"])
             else:
                 self.received_detail += 1
                 self.events.append(event)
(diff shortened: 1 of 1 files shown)


RealCLanger:
Hm, interesting finding. Maybe what you found out is the right approach anyway and we don't need an option for the old matching. I will try this out...

ExploracuriousAlex:
Fixes #183

RealCLanger:
I tested your change and I think matching by subscription id is the only right thing to do. Please remove the new option and only update the code in timeline.py. Then we can merge it. 😄

ExploracuriousAlex:
@RealCLanger Option was removed and code was streamlined.

PR #267, opened 2025-11-12
remove --no-flat option

ExploracuriousAlex:
Better read it twice. 🕵️
I don't want to remove the `--flat` option, but the `--no-flat` option. 😉

Using `action=argparse.BooleanOptionalAction` for the `--flat` option automatically creates both `--flat` and `--no-flat` options.
Since that doesn't make sense in this scenario, better use `action="store_true"`.


RealCLanger (approved):
Good. Makes sense. 😄

PR #275, opened 2025-12-03
Use sys.exit() and import sys

yubiuser:
When exiting, `exit()` should only be used in interactive sessions. Background is that `exit()` needs the `site` module to be loaded. Therefore, use of `syst.exit()` with explicit `sys import` should be preferred.

> They are useful for the interactive interpreter shell and should not be used in programs.

https://docs.python.org/3/library/constants.html#constants-added-by-the-site-module

Also: https://stackoverflow.com/a/19747562

### DIFF
--- a/pytr/__main__.py
+++ b/pytr/__main__.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 import logging
+import sys
 
 from pytr.main import main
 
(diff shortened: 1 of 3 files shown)


RealCLanger:
That's correct. Thanks for your contribution.

Issue #224, opened 2025-06-03
AssertionError while running dl_docs


S-T1:
**Description of the bug**
**While executing "dl_docs" an AssertionError appears in the step "Exporting transactions" "**

**To Reproduce**

1. Ran the command `pytr dl_docs /home/folder1

**Expected behavior**
No error. The data is exported in the mentioned folder1

**Error log**

```
_Exporting transactions ...
Traceback (most recent call last):
  File "/home/stefan/.local/bin/pytr", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/stefan/.local/share/pipx/venvs/pytr/lib/python3.12/site-packages/pytr/main.py", line 354, in main
    asyncio.get_event_loop().run_until_complete(dl.dl_loop())
  File "/usr/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/stefan/.local/share/pipx/venvs/pytr/lib/python3.12/site-packages/pytr/dl.py", line 88, in dl_loop
    self.tl.process_timelineDetail(response, self)
  File "/home/stefan/.local/share/pipx/venvs/pytr/lib/python3.12/site-packages/pytr/timeline.py", line 206, in process_timelineDetail
    self.check_if_done(dl)
  File "/home/stefan/.local/share/pipx/venvs/pytr/lib/python3.12/site-packages/pytr/timeline.py", line 210, in check_if_done
    self.finish_timeline_details(dl)
  File "/home/stefan/.local/share/pipx/venvs/pytr/lib/python3.12/site-packages/pytr/timeline.py", line 237, in finish_timeline_details
    ).export(
      ^^^^^^^
  File "/home/stefan/.local/share/pipx/venvs/pytr/lib/python3.12/site-packages/pytr/transactions.py", line 172, in export
    writer.writerows(transactions)
  File "/usr/lib/python3.12/csv.py", line 167, in writerows
    return self.writer.writerows(map(self._dict_to_list, rowdicts))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stefan/.local/share/pipx/venvs/pytr/lib/python3.12/site-packages/pytr/transactions.py", line 164, in <genexpr>
    transactions = (txn for event in events for txn in self.from_event(event))
[...]

PR #311, opened 2026-02-07
Adjust value assignment logic in event.py

M123-dev:
Fixes: #224

Somehow I (and some others) got a "Verkaufsorder"/Sale with the attribute
```
"amount": {
      "currency": "EUR",
      "value": 0.0,
      "fractionDigits": 2
    },
```
Which caused an error because the parser explicitly checked for `value != 0.0`.

This pull request removes the explicit check for zero values when parsing the `amount.value` field, so a value of `0.0` will now be considered valid instead of being replaced with `None`.

-----
That being said I am not sure why there was this explicit check, if it has a specific reason this PR should probably not be merged as it. I am using my fork right now, and it works for me 👍🏼

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Mypy type check
         run: uv run mypy .
       - name: Check if README.md automatically generated content is up to date
-        run: diff -u -Z README.md <(uvx mksync@0.1.5 README.md)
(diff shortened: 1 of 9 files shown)


RealCLanger:
Thanks for the contribution. I enabled the checks and it shows that there is a formatting error. Can you please take a look and update your change.

The parsing tests seem to work fine, though, so I think your change should be integrated when everything is green.

M123-dev:
Heyy @RealCLanger thanks for the review. I have updated the PR so that the linter should be satisfied.

RealCLanger:
Looks good now, I also tested the change with my data and it shows no regressions. Merging...

Issue #317, opened 2026-03-15
Tax sign is flipped for old dividend taxes


jannewulf:
**Description of the bug**
Dividend taxes have been reported in a different format in the past. The old format had a flipped sign on the taxes. This is not handled in the parsing code. 

According to my data this format change was implemented somewhere between 2024-05-17 and 2024-06-28.

This leads to inconsistent signing of the tax values.

**To Reproduce**
These are samples from my own data:

**2024-05-17**
```json
{
  "title": "Steuern",
  "detail": {
    "text": "0,60 €",
    "trend": null,
    "action": null,
    "displayValue": null,
    "type": "text"
  },
  "style": "plain"
}
```

**2024-06-28**
```json
{
  "title": "Steuer",
  "detail": {
    "text": "-0,08 €",
    "displayValue": {
      "text": "-0,08 €"
    },
    "type": "text"
  },
  "style": "plain"
}
```

**Expected behavior**
The sign on the tax data in the export should be consistent.

**Additional context**

jannewulf:
I've supplied a MR that fixes this issue

RealCLanger:
I'd like to fix this with some larger code cleanup/improvement in #344

RealCLanger:
PR merged.

PR #318, opened 2026-03-15
Fix tax sign in old style events

jannewulf:
This fixes issue #317 

If you look at the surrounding tests, you'll see that the one I changed didn't match the pattern of the other tests.

Please verify:
- with your own data
- that the handling is implemented in the correct location

Note: this is a breaking change to the existing behavior


RealCLanger:
Please see #344

jannewulf:
Thanks, sounds good. Feel free to close this PR

PR #320, opened 2026-03-24
Fix aws-waf

### LINKED ISSUE #319: Login fails with "Expecting value: line 1 column 1 (char 0)"
**Description of the bug**
Login fails:
2026-03-16 10:41:56+0100 main      DEBUG    logging is set to debug
2026-03-16 10:41:56+0100 account   INFO     Found credentials file
2026-03-16 10:41:56+0100 account   INFO     Phone: +491**********, PIN: ****
2026-03-16 10:41:57+0100 account   CRITICAL Expecting value: line 1 column 1 (char 0) 

**To Reproduce**
pytr login

**Expected behavior**
Login works ;-)

**Error log**

**Environment**

- OS: Mac
- pytr version: 0.4.6
 
**Additional context**
It was working approx. a week ago without issues
I deleted credentials and cookie files but that didn't help either (login fails after entering the PIN)

Felixoid:
The work is done on top of @Rick7221's commit from https://github.com/pytr-org/pytr/issues/319#issuecomment-4098802460

- Vendor https://github.com/xKiian/awswaf into the repository since it is:
     - not available in PyPI
     - Depends on unmaintained pyscrypt (https://github.com/xKiian/awswaf/pull/4)
     - Does not provide the resolver for `mp_verify`, AKA network bandwidth challenge
- Implement WAF challenge resolution
- Add optional dependencies

fixes #319

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,8 @@ dependencies = [
     "babel",
     "certifi",
     "coloredlogs",
+    "cryptography",
+    "curl_cffi",
     "ecdsa",
     "packaging",
     "pathvalidate",
--- a/pytr/account.py
+++ b/pytr/account.py
@@ -18,7 +18,7 @@ def get_settings(tr):
         return formatted_json
 
 
-def login(phone_no=None, pin=None, web=True, store_credentials=False):
+def login(phone_no=None, pin=None, web=True, store_credentials=False, waf_token=None):
     """
     If web is true, use web login method, else simulate app login.
     Handle credentials parameters and store to credentials file if requested.
@@ -57,7 +57,7 @@ def login(phone_no=None, pin=None, web=True, store_credentials=False):
         else:
             save_cookies = False
 
-    tr = TradeRepublicApi(phone_no=phone_no, pin=pin, save_cookies=save_cookies)
+    tr = TradeRepublicApi(phone_no=phone_no, pin=pin, save_cookies=save_cookies, waf_token=waf_token)
 
     if web:
         # Use same login as app.traderepublic.com
--- a/pytr/api.py
+++ b/pytr/api.py
@@ -25,6 +25,7 @@ import base64
 import hashlib
 import json
 import pathlib
+import re
 import ssl
 import time
 import urllib.parse
@@ -220,9 +221,14 @@ class TradeRepublicApi:
         try:
             session = cffi_requests.Session(impersonate="chrome")
             response = session.get(self._waf_login_url)
-            host = response.text.split('src="https://')[1].split("/challenge.js")[0]
-            challenge_js = session.get(f"https://{host}/challenge.js").text
-            token = AwsWaf(host, "app.traderepublic.com", challenge_js)()
+            m = re.search(r'src="(https://[^"]+/challenge\.js)"', response.text)
+            if not m:
+                self.log.warning("challenge.js URL not found in login page")
+                return None
+            challenge_js_url = m.group(1)
+            waf_endpoint = challenge_js_url.split("https://", 1)[1].rsplit("/challenge.js", 1)[0]
+            challenge_js = session.get(challenge_js_url).text
+            token = AwsWaf(waf_endpoint, "app.traderepublic.com", challenge_js)()
             self.log.info("AWS WAF token obtained automatically")
             return token
         except Exception as e:
--- a/pytr/api.py
+++ b/pytr/api.py
@@ -221,7 +221,8 @@ class TradeRepublicApi:
             session = cffi_requests.Session(impersonate="chrome")
             response = session.get(self._waf_login_url)
             host = response.text.split('src="https://')[1].split("/challenge.js")[0]
-            token = AwsWaf({}, host, "app.traderepublic.com")()
+            challenge_js = session.get(f"https://{host}/challenge.js").text
+            token = AwsWaf(host, "app.traderepublic.com", challenge_js)()
             self.log.info("AWS WAF token obtained automatically")
             return token
         except Exception as e:
@@ -246,7 +247,7 @@ class TradeRepublicApi:
             discard=False,
             comment=None,
             comment_url=None,
-            rest={"HttpOnly": None},
+            rest={"HttpOnly": ""},
         )
         self._websession.cookies.set_cookie(cookie)
 
--- a/pytr/api.py
+++ b/pytr/api.py
@@ -286,7 +286,12 @@ class TradeRepublicApi:
     def save_websession(self):
         # Saves session cookies too (expirydate=0).
         if self._save_cookies:
-            self._websession.cookies.save(ignore_discard=True, ignore_expires=True)
+            # Save a copy without the WAF token - it's fetched fresh on every startup
+            save_jar = MozillaCookieJar(self._cookies_file)
+            for cookie in self._websession.cookies:
+                if cookie.name != "aws-waf-token":
+                    save_jar.set_cookie(cookie)
+            save_jar.save(ignore_discard=True, ignore_expires=True)
 
     def resume_websession(self):
         """
@@ -301,11 +306,14 @@ class TradeRepublicApi:
             # Loads session cookies too (expirydate=0).
             self._websession.cookies.load(ignore_discard=True, ignore_expires=True)
             self._weblogin = True
+            # Re-apply fresh WAF token over any stale one from the cookie file
+            if self._waf_token:
+                self._set_waf_cookie(self._waf_token)
             try:
                 self.settings()
             except requests.exceptions.HTTPError:
-                return False
                 self._weblogin = False
+                return False
             else:
                 return True
         return False
--- a/pytr/api.py
+++ b/pytr/api.py
@@ -29,7 +29,7 @@ import ssl
 import time
 import urllib.parse
 import uuid
-from http.cookiejar import MozillaCookieJar
+from http.cookiejar import Cookie, MozillaCookieJar
 from typing import Any, Dict
 
 import certifi
@@ -92,6 +92,7 @@ class TradeRepublicApi:
         save_cookies=False,
         credentials_file=None,
         cookies_file=None,
+        waf_token=None,
     ):
         self.log = get_logger(__name__)
         self._locale = locale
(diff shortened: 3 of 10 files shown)


kugel-:
I tried this and it worked. Many great thanks for that!

RealCLanger:
I tried it and it worked once. But after that it doesn't any more. Trying to figure out what happens...

Felixoid:
Should be cookies in `~/.pytr`

Probably, worth cleaning them on 403?

RealCLanger:
Hm, ok, interesting. But shouldn't the cookies be updated, at least the aws-waf-token which is the one that counts. I'll try/check that.

Update: Cleaning the cookies helped.

Felixoid:
I manually deleted the `aws-waf-token` cookie from a file, and it worked too.

Probably, we need to filter it when reading from the cookie file

RealCLanger:
Generally, the change looks very good. Thanks for the contribution. I'll try to review it more thoroughly later today.

Felixoid:
Thanks!

I added a fix. We clearly don't need to save the WAF cookie. And if there's one in the file, we need to overwrite it with the fresh one.

RealCLanger:
> Thanks!
> 
> I added a fix. We clearly don't need to save the WAF cookie. And if there's one in the file, we need to overwrite it with the fresh one.

Maybe the WAF token could be stored, though and used. And only if we get a 403 in authentication we update/regenerate it. I guess that would at least improve performance.

I don't know whether there's (another) way to check whether the waf token is still valid?

MedAzizKhayati:
They changed the login yet again it seems, I am getting 405 response code when I try to login via API.

RealCLanger:
> They changed the login yet again it seems, I am getting 405 response code when I try to login via API.

I saw it failing again and needed to remove the cookies once more, did you try that? I also didn't add the latest commit yet to my build that filters/updates the WAF cookie.

Felixoid:
> I saw it failing again and needed to remove the cookies once more

Didn't affect me with the latest code, works after 2h of the last launch

> Maybe the WAF token could be stored, though and used. And only if we get a 403 in authentication we update/regenerate it. I guess that would at least improve performance.

It's a subsecond. I think automatically obtained token is a safe side:

```
In [8]: def bench():
   ...:     time_start = time.perf_counter()
   ...:     i = 50
   ...:     for _ in range(i):
   ...:         api.TradeRepublicApi()
   ...:     time_end = time.perf_counter()
   ...:     time_duration = time_end - time_start
   ...:     avg_time = time_duration / i
   ...:     print(f'Took {time_duration:.3f} seconds, average time {avg_time:.3f} seconds')
   ...:

In [9]: bench()
12:08:30 AWS WAF token obtained automatically
12:08:30 AWS WAF token obtained automatically
12:08:30 AWS WAF token obtained automatically
12:08:31 AWS WAF token obtained automatically
12:08:31 AWS WAF token obtained automatically
12:08:31 AWS WAF token obtained automatically
12:08:32 AWS WAF token obtained automatically
12:08:32 AWS WAF token obtained automatically
12:08:32 AWS WAF token obtained automatically
12:08:32 AWS WAF token obtained automatically
12:08:33 AWS WAF token obtained automatically
12:08:33 AWS WAF token obtained automatically
12:08:33 AWS WAF token obtained automatically
12:08:34 AWS WAF token obtained automatically
12:08:34 AWS WAF token obtained automatically
12:08:34 AWS WAF token obtained automatically
12:08:35 AWS WAF token obtained automatically
12:08:35 AWS WAF token obtained automatically
12:08:35 AWS WAF token obtained automatically
12:08:35 AWS WAF token obtained automatically
12:08:36 AWS WAF token obtained automatically
12:08:36 AWS WAF token obtained automatically
12:08:36 AWS WAF token obtained automatically
12:08:37 AWS WAF token obtained automatically
12:08:37 AWS WAF token obtained automatically
12:08:37 AWS WAF token obtained automatically
12:08:37 AWS WAF token obtained automatically
12:08:38 AWS WAF token obtained automatically
12:08:38 AWS WAF token obtained automatically
12:08:38 AWS WAF token obtained automatically
12:08:38 AWS WAF token obtained automatically
12:08:39 AWS WAF token obtained automatically
12:08:39 AWS WAF token obtained automatically
12:08:39 AWS WAF token obtained automatically
12:08:40 AWS WAF token obtained automatically
12:08:40 AWS WAF token obtained automatically
12:08:40 AWS WAF token obtained automatically
12:08:41 AWS WAF token obtained automatically
12:08:41 AWS WAF token obtained automatically
12:08:41 AWS WAF token obtained automatically
12:08:41 AWS WAF token obtained automatically
12:08:42 AWS WAF token obtained automatically
12:08:42 AWS WAF token obtained automatically
12:08:42 AWS WAF token obtained automatically
12:08:43 AWS WAF token obtained automatically
12:08:43 AWS WAF token obtained automatically
12:08:43 AWS WAF token obtained automatically
12:08:43 AWS WAF token obtained automatically
12:08:44 AWS WAF token obtained automatically
12:08:44 AWS WAF token obtained automatically
Took 14.585 seconds, average time 0.292 seconds
```

RealCLanger:
> > I saw it failing again and needed to remove the cookies once more
> 
> Didn't affect me with the latest code, works after 2h of the last launch

Yes, I had the error again, then updated to the latest version and it got resolved.

> > Maybe the WAF token could be stored, though and used. And only if we get a 403 in authentication we update/regenerate it. I guess that would at least improve performance.
> 
> It's a subsecond. I think automatically obtained token is a safe side

Probably yes, if we don't have a client-side mechanism to check validity of the token.

tobias-goertz:
@Felixoid Great work on this PR — tested it and the WAF token generation works perfectly.

One issue I ran into: the WAF token is fetched using `curl_cffi` (browser TLS fingerprint), but the actual login POST to `/api/v1/auth/web/login` still goes through plain `requests.Session`. TR's ALB appears to validate TLS fingerprints on auth endpoints — I consistently get 405 from `awselb/2.0` with plain `requests`, even with a valid WAF token.

Replacing `_websession` with a `curl_cffi.Session(impersonate="chrome")` for the auth requests fixes this. Since `curl_cffi` is already a dependency of this PR, it would just be a matter of using it for `_websession` too (or at least for the auth-related HTTP calls).

This might explain why some people in #319 report 405 errors even after applying this fix — it depends on whether TR's ALB enforces TLS fingerprinting for their specific requests/region.

Also minor: the WAF host extraction via `response.text.split('src="https://')[1].split("/challenge.js")[0]` is a bit fragile — a regex like `re.search(r'src="(https://[^"]+)/challenge\.js"', response.text)` would be more robust against HTML changes.

Felixoid:
> Replacing `_websession` with a `curl_cffi.Session(impersonate="chrome")` for the auth requests fixes this. Since `curl_cffi` is already a dependency of this PR, it would just be a matter of using it for `_websession` too (or at least for the auth-related HTTP calls).

Currently, it's an optional dependency. 

I can think of conditional import in this place, so if a user has curl_cffi installed, it will be used. 

Although, for a manual set token it's a slippery slope, since an original browser can be anything. 

I don't think an issue is a session. Most probably it's an outdated cookie from the file with previous broken code

Felixoid:
> the WAF host extraction via response.text.split('src="https://')[1].split("/challenge.js")[0] is a bit fragile — a regex like re.search(r'src="(https://[^"]+)/challenge\.js"', response.text) would be more robust against HTML changes.

After `AwsWaf` started parsing challenge.js, the logic should be twisted around, as you suggested. Thanks, it's much more reliable indeed!

Felixoid:
> This might explain why some people in https://github.com/pytr-org/pytr/issues/319 report 405 errors even after applying this fix — it depends on whether TR's ALB enforces TLS fingerprinting for their specific requests/region.

There was a state, when `aws-waf-token` cookie was written into a cookie jar on script exit. Then, on another run a new token was received, but after that, the cookies were restored from the file, overwriting `aws-waf-token` with an outdated value

It was fixed in c81d47f3ce1db07372fd9d4bcc883c36e1799d53, please test it

tobias-goertz:
@Felixoid You were right — the 405 I was seeing was caused by stale WAF cookies, not TLS fingerprinting. After updating to your latest commits (cookie cleanup fix), I removed my `curl_cffi` BrowserSession workaround entirely and everything works with plain `requests`.

Tested the full flow end-to-end: WAF token → login → 2FA → websocket data sync. All good on latest `a9f6116`.

Thanks for the quick iteration on the cookie handling!

RealCLanger:
A formal remark: Could you put the MIT license first in the vendored files and the comment from where it was vendored etc. afterwards?

Also, I suggest to remove the additional LICENSE file in the aswwaf subdirectory as it is the same license (MIT) as for the other files and additionally, it is included as file header.

Felixoid:
Let me rebase the staff a little bit, so it will be cleaner git log.

The work looks finished.

RealCLanger:
> Let me rebase the staff a little bit, so it will be cleaner git log.
> 
> The work looks finished.

We'll squash anyway 😄

RealCLanger (commented):
OK, this looks close to merging now. Two small remarks in the code.

RealCLanger:
Thanks @Felixoid for this great piece of work.

fft1010:
Hi. I can confirm, that it now works for non programmers as me. Thanks a lot for this. But if I do
`pytr -V`
it said, I am on the latest version, doing a 
```
pip install --upgrade --force-reinstall git+https://github.com/pytr-org/pytr.git
```
got the newest version, and it worked. Again. Thanks a lot, this is for others, that might not find the "trick" on their own.

Juergen

RealCLanger:
> Hi. I can confirm, that it now works for non programmers as me. Thanks a lot for this. But if I do `pytr -V` it said, I am on the latest version, doing a
> 
> ```
> pip install --upgrade --force-reinstall git+https://github.com/pytr-org/pytr.git
> ```
> 
> got the newest version, and it worked. Again. Thanks a lot, this is for others, that might not find the "trick" on their own.

Version 0.4.7 with the fix is now officially released and available via standard `pip upgrade`.

xKiian:
lol

RealCLanger:
> lol
?

Looks like aws-waf does not work for the TR API at the moment.

@xKiian

xKiian:
Sorry but I couldn't bother fixing it. People are reselling my code for 5figs...

PR #329, opened 2026-04-04
Add TAX_YEAR_END_REPORT_CREATED to known events

zimmermq:
New Trade Republic event type for annual tax report ("Steuerbericht"). Added to events_known_ignored and event_subfolder_mapping so the PDF is downloaded into Misc/ without triggering an unknown event warning.

Fixes https://github.com/pytr-org/pytr/issues/328


RealCLanger (approved):
Looks good. Thanks for the contribution.

PR #344, opened 2026-04-20
Taxes and Fees should always be positive numbers

RealCLanger:
There is some inconsistency with the sign of tax values. It should always be positive. Also, fees should have a positive number.

This is a breaking change, that is, modifying current behavior that might be relied on elsewhere.

With the patch we also go another step towards unification of event parsing in one method. This can help to simplify it in the end.

It addresses #317.

### DIFF
--- a/pytr/event.py
+++ b/pytr/event.py
@@ -299,8 +299,33 @@ class Event:
         note: Optional[str] = None
         subtitle = event_dict["subtitle"]
         eventdesc = f"{title} {subtitle} ({event_dict['id']})"
+        dump_dict = {"eventdesc": eventdesc, "id": event_dict["id"]}
+        pref_locale = "de"
+        (
+            fees_dict,
+            taxes_dict,
+        ) = (None,) * 2
+
         sections = event_dict.get("details", {}).get("sections", [{}])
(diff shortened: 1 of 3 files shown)


pytr/event.py
```diff
@@ -444,6 +469,39 @@ def from_dict(cls, event_dict: Dict[Any, Any]):
...
+            get_event_logger().debug("Failed to parse fees from %s", json.dumps(event_dict, indent=4))
+
+        # parse taxes
+        if taxes_dict:
+            dump_dict["subtitle"] = taxes_dict["title"]
+            dump_dict["type"] = "taxes"
+            taxes = cls._parse_float_from_text_value(
+                taxes_dict.get("detail", {}).get("text", ""), dump_dict, pref_locale
+            )
+            if taxes and taxes < 0:
```
jannewulf, line 502: Did you check that it's working correctly with tax refunds as well?

RealCLanger: > Did you check that it's working correctly with tax refunds as well?

Yes, it is tested explicitly here:
https://github.com/RealCLanger/pytr/blob/70b2106e8c2e13e9f903ab379c19da0504d065d3/tests/test_events.py#L1529

PR #355, opened 2026-05-29
Document the --v2 web login flag in the README

### LINKED ISSUE #250: App Login Broken
**Description of the bug**
Today I tried to fetch data from trade republic and it seems the app login is broken. 
The initiate_device_reset method is not working.
From the error message it seems something is outdated on the client side. 

**What I tried already**
- I already tried to update all the python libraries (certifi, websockets, requests etc.).
- I randomly tried different User-Agent headers. Nothing worked.
- I tried to reverse engineer the User-Agent headers using by my phone through a http proxy, but it seems trade republic blocks the CA certificate.

**To Reproduce**
Minimum code example:

```
import hashlib
from ecdsa import NIST256p, SigningKey
import requests

_default_headers = {"User-Agent": "TradeRepublic/Android 30/App Version 1.1.5534"}
_host = "https://api.traderepublic.com"

sk = SigningKey.generate(curve=NIST256p, hashfunc=hashlib.sha512)
r = requests.post(
    f"{_host}/api/v1/auth/account/reset/device",
    json={"phoneNumber": phone_number, "pin": pin},
    headers=_default_headers,
)
print(r.status_code)
print(r.json())
```

**Expected behavior**
This code should trigger a token to be send.

**Error log**
```
status code: 426
{'errors': [{'errorCode': 'CLIENT_VERSION_OUTDATED', 'errorMessage': None, 'meta': None}]}
```

**Environment**

- OS: windows
- package manager: conda
- python: 3.13.9
- pytr version: 0.4.3

twannooitmeer:
> ## Update, 2026-08-06: scope reduced to documentation only
>
> Since this PR was opened, #380 ("New login method") landed on `master` and implements
> essentially the same v2 web-login change proposed here, and does it better. Rather than
> leave a conflicting duplicate open, I have merged `master` in and dropped my
> implementation entirely: `pytr/api.py`, `pytr/account.py` and `pytr/main.py` on this
> branch are now byte-identical to `master`.
>
> Where #380 improves on what I wrote:
>
> * it handles accounts secured with an authenticator app
>   (`requiredAction == AUTHENTICATOR_VERIFICATION`), which my version did not;
> * it derives the approval-poll deadline from the login process itself instead of my
>   hard-coded 180s timeout;
> * it maps login error codes to readable messages and keeps the process id out of them;
> * it correctly reports that Trade Republic removed the SMS resend endpoint along with
>   the v1 web login. My version still posted to a v2 resend path that does not exist;
> * it ships `tests/test_api_urls.py`. My version had no tests.
>
> **What is left in this PR is the README section documenting the `--v2` flag**, which
> #380 did not add. The flag is currently undocumented, so a user has no way to discover
> it other than reading `--help`. The text describes #380's behaviour, not mine: it covers
> the authenticator-app path and notes there is no SMS fallback under `--v2`.
>
> I have also dropped the `Closes #250` keyword from this PR, since #380 is what actually
> resolved that issue. Happy to close this entirely instead if you would rather take the
> README separately or not at all.
>
> The original description follows for context.

---

<details>
<summary>Original description (superseded by #380)</summary>

## Summary

Trade Republic deprecated the legacy `/api/v1/auth/web/login` endpoint. All requests to it currently return `HTTP 426 CLIENT_VERSION_OUTDATED` regardless of `User-Agent`, which makes `pytr login` unusable on a fresh install (refs #250). (Verified on 2026-05-29; the cutover itself happened at some earlier point.)

This PR migrates the auth flow to `/api/v2/auth/web/login`, which is what `app.traderepublic.com` currently uses.

### `/api/v2/auth/web/login` requires four new headers

Captured from the live web app:

```
x-tr-platform: web
x-tr-app-version: 15.7.0          # web track, distinct from the Android app version
x-tr-device-info: <base64(JSON)>  # browser fingerprint
x-aws-waf-token: <same value as the aws-waf-token cookie>
```

The `x-tr-device-info` payload looks like:

```json
{
  "stableDeviceId": "<64 hex chars>",
  "model": "Apple Macintosh",
  "browser": "Chrome",
  "browserVersion": "148.0.0.0",
  "os": "Mac OS",
  "osVersion": "10.15.7",
  "timezone": "Europe/Amsterdam",
  "timezoneOffset": -120,
  "screen": "1800x1169x30",
  "preferredLanguages": ["en", "en-US"],
  "numberOfCores": 12,
  "deviceMemory": 16
}
```

### v2 uses push-approval, not a numeric code

The v2 flow no longer issues an SMS/numeric code. Instead the user approves the login via a push notification in the Trade Republic mobile app, and the web client polls `GET /api/v2/auth/web/login/processes/{processId}` until the response `status` indicates approval (and a `tr_session` cookie is set).

## Implementation

- `pytr/api.py`
  - Module-level constants `TR_WEB_APP_VERSION`, `TR_WEB_USER_AGENT`, `TR_WEB_LOGIN_PATH`, overridable via `PYTR_TR_APP_VERSION` / `PYTR_TR_USER_AGENT` so future TR bumps don't require a code change:
    ```bash
    # If TR bumps the web app version and pytr starts returning CLIENT_VERSION_OUTDATED again:
    PYTR_TR_APP_VERSION=15.8.0 pytr login

    # If TR starts gating by a specific browser UA:
    PYTR_TR_USER_AGENT='Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/149.0.0.0 Safari/537.36' pytr login
    ```
  - `_get_device_id()` persists a stable 64-hex device id at `~/.pytr/device_id`.
  - `_build_device_info_header()` constructs the base64-encoded JSON fingerprint.
  - `_auth_headers()` returns all four required headers (WAF token included when known).
  - `await_web_login_approval()` polls the new `processes/{processId}` endpoint every 2 s with a 180 s timeout. Detects approval via either an explicit `status` field or the presence of a `tr_session` cookie.
  - `initiate_weblogin()` now POSTs to v2 with the new headers, kicks off the poll, and returns `0` to signal the v2 flow.
  - `complete_weblogin('')` is a no-op (approval already happened during initiation).
- `pytr/account.py`
  - When `initiate_weblogin()` returns `0`, skip the code/SMS prompts entirely and call `complete_weblogin('')` to persist cookies.
  - Legacy v1 prompt path is preserved unchanged for the `countdown > 0` case (dead branch in practice, but safe to keep).

## Tested

Verified locally end-to-end on macOS, Python 3.12: WAF token fetch → `processId` → push approval → session cookie → `Logged in.`

## Notes

- Does not yet implement opt-in registration via `trustDeviceRegistrationOptions` (returned by the process endpoint); that would let users skip the push step on subsequent logins. Happy to follow up in a separate PR.

</details>


MedAzizKhayati:
Worked for me!

RealCLanger:
Thanks for the contribution. As the current web login method still works, would you mind keeping the coding available as well and add an option for selecting the desired one?

twannooitmeer:
@RealCLanger i can look into this sometime soon. For me the "current" sign in method was not working as my TR account only allows for sign in confirmation through in-app notification. I was not able to get TR to text me a code (hence the reworked method).

How do you suggest we implement the choice?

MedAzizKhayati:
> How do you suggest we implement the choice?

I would suggest adding an option `pytr login --v2`

twannooitmeer:
I made the requested change(s)

- pytr login was reverted back to main and keeps the legacy flow exactly as before (same endpoint, same headers, same SMS/code prompt).
- TradeRepublicApi default constructor keeps identical behaviour to current master
- pytr login --v2 opts into the new push-approval flow. The flag lives on the shared _parse_login_args_ so it works for every subcommand  that also executes login (pytr portfolio --v2, pytr dl_docs --v2, etc.) for users whose accounts can't get a numeric code.
- All the v2-specific config is gated behind _use_v2_login=True_ on the API object, so it can't affect v1 callers.
- I also included the --v2 flag in the readme of the repo

As I do not have a TR account that can verify the legacy/v1 flow, I did not do an end to end test for that version. Ruff check and pytest were succesful. I also did a test of the --v2 flow end-to-end. This works for me as well.

RealCLanger:
Thanks for updating the PR. As you've seen I have merged the code from #373 but added the flag to switch between login methods. I also saw that still the documentation should be updated, so thanks for doing that.

PR #365, opened 2026-07-18
Fix crash in _parse_card_note when eventType is null

Titouan-Loeb:
event_dict.get("eventType", "") only falls back to "" when the key is missing, not when it is present but null, which Trade Republic sends for some card events. That None then hits .startswith("card_") and raises AttributeError: 'NoneType' object has no attribute 'startswith'.

Added a regression fixture with eventType explicitly null (the existing kartenzahlung_no_eventType.json only covers the key being absent, a different case)

### DIFF
--- a/pytr/event.py
+++ b/pytr/event.py
@@ -765,7 +765,7 @@ class Event:
             Optional[str]: note
         """
         eventTypeStr = event_dict.get("eventType", "")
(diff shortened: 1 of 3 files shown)


RealCLanger:
Makes sense. Thanks for the fix.

PR #368, opened 2026-07-25
Store credentials file with restricted 600 perms

KevSlashNull:
This adds a `chmod` that changes the permissions of the credentials file
`~/.pytr/credentials` to only allow reading and writing by the user and not the
group or other users.

SSH does something similar where it requires that private keys must be set to
600 instead of the default 644 mode.

### DIFF
--- a/pytr/account.py
+++ b/pytr/account.py
@@ -1,4 +1,5 @@
 import json
+import os
 import sys
 import time
 from getpass import getpass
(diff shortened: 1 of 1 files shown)


RealCLanger (changes requested):
Makes sense but the change still has an error. See checks. Please update and fix.

KevSlashNull:
ruff, oops. ruff checks pass now, thanks @RealCLanger.

RealCLanger (approved):
Looks good now. Thanks for the contribution.

PR #387, opened 2026-08-08
Classify negative cash invoices as taxes

### LINKED ISSUE #163: Vorabpauschale gets parsed as Dividend
I just got the recent Vorabpauschale transactions and they were mapped to Dividends (expected type would be `Taxes`).
It uses the same eventType `ssp_corporate_action_invoice_cash`, but the Amount is negative this time.

Here's a full transaction json example:
[vorabpauschale.json](https://github.com/user-attachments/files/18548058/vorabpauschale.json)

Guflly:
Fixes #163.

Negative `SSP_CORPORATE_ACTION_INVOICE_CASH` events are now classified as taxes before applying the dividend subtitle fallback. Positive, zero, and missing amounts keep their existing behavior, and canceled events remain ignored.

Tests:
- `uv run pytest -q --basetemp .venv/pytest-temp`
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run mypy .`

### DIFF
--- a/pytr/event.py
+++ b/pytr/event.py
@@ -469,12 +469,10 @@ class Event:
             elif eventTypeStr in ["SSP_CORPORATE_ACTION_INVOICE_CASH", "SSP_CORPORATE_ACTION_CASH"]:
                 if subtitle == "Aufruf von Zwischenpapieren":
                     event_type = PPEventType.SWAP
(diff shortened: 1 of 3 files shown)


RealCLanger:
Thanks for proposing this code simplification. I even made it more concise. Regarding the tests you added, I think most of that stuff is covered anyway in test_events.py. I've added another fixture for a "Vorabpauschale" to test that it is correctly parsed.

